PR #6, opened 2024-07-12
Convert from React router to tanstack router

djnunez-aot:


submit-web/.gitignore
```diff
@@ -11,6 +11,7 @@ node_modules
 dist
 dist-ssr
 *.local
+package-lock.json
```
jadmnsaad: it's okay if we commit this, let's not ignore it

PR #17, opened 2024-07-26
Create tables and migrations

jadmnsaad:
Following the result of https://docs.google.com/spreadsheets/d/1E1EH5Sq2W9Y0Qwpjbo6va7W9VTE_RNu9M5_pTCo4DoY/edit?gid=0#gid=0 from https://apps.nrs.gov.bc.ca/int/jira/browse/EPICSYSTEM-206

I created the tables and the migrations


submit-api/migrations/versions/0223afc14044_.py
```diff
@@ -0,0 +1,101 @@
...
+# revision identifiers, used by Alembic.
+revision = '0223afc14044'
+down_revision = '20bcb68bc2ac'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('account',
```
dinesh-aot: Should the table names be singular or plural

submit-api/src/submit_api/models/role.py
```diff
@@ -0,0 +1,19 @@
...
+from .db import db
+
+
+class Role(BaseModel):
+    """Definition of the Role entity."""
+
+    __tablename__ = 'role'
+
+    id = Column(db.Integer, primary_key=True, autoincrement=True)
+    role_name = Column(db.String(50), nullable=False)
```
saravanpa-aot, line 19: whats the role like ? not the keycloak role rite? can you add more comments here plz...

submit-api/migrations/versions/0223afc14044_.py
jadmnsaad: changed to plural

saravanpa-aot (approved):

PR #19, opened 2024-07-29
Two level link setup - Jira EPIC SYSTEM 205

djnunez-aot:
-Setup sub routes for side nav

-Add hex to rgb function 

image


submit-web/src/components/Shared/SideNavBar.tsx
```diff
@@ -51,34 +31,79 @@ export default function SideNavBar() {
...
-                  <span style={{ color: "inherit" }}>{route.routeName}</span>
-                </ListItemButton>
-              </Link>
-            </ListItem>
+                  <ListItemButton
+                    sx={{
+                      pl: "2rem",
+                      backgroundColor:
+                        currentPath === route.path
+                          ? hexToRgba(theme.palette.secondary.main, 0.1)
```
jadmnsaad: why do we need this util method?

djnunez-aot: To get the design that was requested we need to apply opacity to the hex values

submit-web/src/styles/utils.ts
```diff
@@ -0,0 +1,13 @@
+export const hexToRgba = (hex: string, opacity: number) => {
```
jadmnsaad: if u don't need this anymore, let's remove it

jadmnsaad (approved):

PR #21, opened 2024-07-30
Add create account CRUD

jadmnsaad:
https://apps.nrs.gov.bc.ca/int/jira/browse/EPICSYSTEM-208

- Add create account resource
- Add request schemas
- Add services
- Add model methods


davidjnunezf (approved):

submit-api/migrations/versions/97de805275ec_.py
```diff
@@ -0,0 +1,56 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.bulk_insert(
+        roles_table,
+        [
+            {
+                'role_name': 'ACCOUNT_PRIMARY_ADMIN',
+                'description': 'Administrator role',
+                'created_date': datetime.utcnow(),
+                'created_by': 'system'  # Or replace with the actual creator
```
dinesh-aot, line 41: What do you think of adding rest of the roles if you already know them like PROJECT_ADMIN

submit-api/src/submit_api/models/account.py
```diff
@@ -16,4 +16,22 @@ class Account(BaseModel):
...
+        """Fetch account by proponent id."""
+        return cls.query.filter_by(proponent_id=proponent_id).first()
+
+    @classmethod
+    def create_account(cls, account_data, session=None) -> Account:
+        """Create account."""
+        account = Account(
+            proponent_id=account_data.get('proponent_id', None),
+        )
+        if session:
```
dinesh-aot, line 32: May I know the purpose of this checking "if session"

jadmnsaad: if we have 3 transactions and one fail and we need to rollback everything if one fails I implemented the above pattern.

you pass the same session for all 3. if no session is passed then it's just db.session.commit or something

submit-api/migrations/versions/97de805275ec_.py
jadmnsaad: let's agree on the list and I can follow that up in another PR

submit-api/src/submit_api/models/account.py
jadmnsaad: image

dinesh-aot: I got it. I hope we will be using it only  in those scenarios where there are multiple database transactions involved.

dinesh-aot (approved):

PR #34, opened 2024-08-19
Header touch ups

djnunez-aot:
List of header changes:
All buttons should have a height of 40px

Reduce height of the image to 56px and reduce the margin on the top and bottom of the image to 16px so total height of the header is reduced to about 88px.

Change the color of the bottom border to #d8d8d8, RGB 216, 216, 216

Add a separator between the logo and EPIC.submit

For @jadmsaadaot :
Change Epic.Submit to EPIC.submit for VITE_APP_TITLE  in env on Openshift

### DIFF
--- a/submit-web/src/components/Shared/layout/Header/EAOAppBar.tsx
+++ b/submit-web/src/components/Shared/layout/Header/EAOAppBar.tsx
@@ -14,7 +14,7 @@ export default function EAOAppBar() {
         position="static"
         color="inherit"
         sx={{
-          borderBottom: `1px solid #d8d8d8`,
+          borderBottom: `1px solid ${BCDesignTokens.surfaceColorBorderDefault}`,
           boxShadow: "none",
         }}
       >
--- a/submit-web/src/components/Shared/layout/Header/EAOAppBar.tsx
+++ b/submit-web/src/components/Shared/layout/Header/EAOAppBar.tsx
@@ -4,7 +4,6 @@ import { AppConfig } from "@/utils/config";
 import AppBarActions from "./AppBarActions";
 import { useIsMobile } from "@/hooks/common";
 import MobileNav from "./MobileNav";
-import { theme } from "@/styles/theme";
 import { BCDesignTokens } from "epic.theme";
 
 export default function EAOAppBar() {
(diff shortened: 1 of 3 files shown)


submit-web/src/components/Shared/layout/Header/EAOAppBar.tsx
```diff
@@ -14,29 +14,31 @@ export default function EAOAppBar() {
         position="static"
         color="inherit"
         sx={{
-          borderBottom: `1px solid ${theme.palette.grey[300]}`,
+          borderBottom: `1px solid #d8d8d8`,
```
jadmnsaad: how can we mitigate using a custom color like this, can we explore using a color that exists in BCDesignTokens?

submit-web/src/styles/theme.tsx
```diff
@@ -1,6 +1,16 @@
...
 
 // Any theme overrides should be passed into the createAppTheme.
 
-export const theme = createAppTheme({});
+export const theme = createAppTheme({
+  components: {
+    MuiButton: {
+      styleOverrides: {
+        root: {
+          height: "40px",
```
jadmnsaad: I'm worried about this, this has caused problems in Track where text has been squished and sometimes it overflowed especially when the text wraps. Maybe also then not allow text to wrap in buttons (some css property)

djnunez-aot: how should we set this up then?

submit-web/src/components/Shared/layout/Header/EAOAppBar.tsx
djnunez-aot: i looked through the bcdesign tokens i did not find d8d8d8

jadmnsaad: then this should prompt a conversation. Let's talk with MA about only using BCDesignTokens

submit-web/src/styles/theme.tsx
jadmnsaad: please look into it or remove this and do it in another PR

jadmnsaad (approved):

PR #35, opened 2024-08-19
Responsiveness Form changes

djnunez-aot:
From MA:
- All the forms on the Entity side need to be fully responsive: the form should remain the same width until the screen gets smaller and then be 100% of the screen.

- The color for validation message should be #CE3E39 (206, 62, 57)

- The validation messages should be “Please enter `<label in lower case>`.” 

- And please replace BC Hydro by CGI Mines Inc. in the first paragraph.

- Another super small thing for the next page (project): Can you change “projects” in the text and on the button for “Project(s)” (keep the lowercase in text and capital on the button but put the “s” in parentheses.)

### DIFF
--- a/submit-web/src/components/Shared/controlled/ControlledTextField.tsx
+++ b/submit-web/src/components/Shared/controlled/ControlledTextField.tsx
@@ -34,7 +34,6 @@ const ControlledTextField: FC<IFormInputProps> = ({
           {...field}
           inputProps={{
             maxLength: maxLength,
-            height: "32px",
           }}
           onChange={(e) => {
             if (onInputChange) {
--- a/submit-web/src/components/Shared/controlled/ControlledTextField.tsx
+++ b/submit-web/src/components/Shared/controlled/ControlledTextField.tsx
@@ -1,6 +1,7 @@
 import React, { FC } from "react";
 import { TextField, TextFieldProps } from "@mui/material";
 import { Controller, useFormContext } from "react-hook-form";
+import { BCDesignTokens } from "epic.theme";
 
 type IFormInputProps = {
   name: string;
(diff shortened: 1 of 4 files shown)


submit-web/src/components/Shared/controlled/ControlledTextField.tsx
```diff
@@ -44,7 +46,11 @@ const ControlledTextField: FC<IFormInputProps> = ({
             field.onChange(e.target.value);
           }}
           error={!!errors[name]}
-          helperText={String(errors[name]?.message ?? "")}
+          helperText={
```
jadmnsaad: use FormHelperTextProps instead

submit-web/src/routes/_authenticated/registration/create-account.tsx
```diff
@@ -56,6 +59,9 @@ function CreateAccount() {
   const methods = useForm({
     resolver: yupResolver(createAccountSchema),
     mode: "onBlur",
+    defaultValues: {
+      phone: "(000) 000-0000",
```
jadmnsaad: I think we should do something else here, why do we need to set the default value like that, can't we set a placeholder in the textfield?

jadmnsaad (changes requested):
- helperText change
- default value -> placeholder instead

submit-web/src/routes/_authenticated/registration/create-account.tsx
djnunez-aot: ill do that

submit-web/src/components/Shared/controlled/ControlledTextField.tsx
```diff
@@ -33,6 +34,7 @@ const ControlledTextField: FC<IFormInputProps> = ({
           {...field}
           inputProps={{
             maxLength: maxLength,
+            height: "32px",
```
jadmnsaad: if we have to do this, please do it at the theme level

jadmnsaad (approved):

PR #40, opened 2024-08-23
Fix projects table spacing

djnunez-aot:
-adjusted padding to table headers

-added more width to the submission name table cell to reduce text overflow on project title(usually submission names are a bit longer)

### DIFF
--- a/submit-web/src/components/Projects/SubmissionTable.tsx
+++ b/submit-web/src/components/Projects/SubmissionTable.tsx
@@ -48,7 +48,7 @@ export default function SubmissionTable({
                 Submitted By
               </TableCell>
               <TableCell
-                align="right"
+                align="center"
                 sx={{ color: BCDesignTokens.themeGray70 }}
               >
                 Status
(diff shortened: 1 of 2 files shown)


jadmnsaad (approved):

submit-web/src/components/Projects/SubmissionTable.tsx
```diff
@@ -27,17 +27,22 @@ export default function SubmissionTable({
         {!headless && (
           <TableHead sx={{ border: 0 }}>
             <TableRow>
-              <TableCell sx={{ color: BCDesignTokens.themeGray70 }}>
+              <TableCell
+                sx={{
+                  color: BCDesignTokens.themeGray70,
+                  width: "30%",
```
jadmnsaad: do we have to manually set the width like this, is there no prop in Table cell that handles this like 
image

jadmnsaad: Is there nothing like a column span sort of prop that can be used?

jadmnsaad (changes requested):
Cell width, better way?

jadmnsaad (approved):

PR #59, opened 2024-09-05
Clean up Submission Id page

djnunez-aot:
-Setup statuses for submissions
-Setup sorting for the table
-Clean up styles to match design
-Add missing labels
-Setup table sorting methods

### DIFF
--- a/submit-web/src/components/Submission/DocumentTable.tsx
+++ b/submit-web/src/components/Submission/DocumentTable.tsx
@@ -88,7 +88,7 @@ export default function DocumentTable({
         </TableHead>
         <TableBody>
           {sortedDocuments?.map((document) => (
-            <CustomRow key={document.id} document={document} />
+            <CustomRow key={`custom-row-${document.id}`} document={document} />
           ))}
           {sortedDocuments.length === 0 && (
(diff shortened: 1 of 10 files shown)


submit-web/src/components/Submission/DocumentTable.tsx
```diff
@@ -71,92 +87,119 @@ export default function DocumentTable({
...
-                sx={{
-                  borderTop: "2px solid #F2F2F2",
-                  borderBottom: "2px solid #F2F2F2",
-                  borderLeft: "2px solid #F2F2F2",
-                  borderTopLeftRadius: 5,
-                  borderBottomLeftRadius: 5,
-                  py: BCDesignTokens.layoutPaddingSmall,
-                }}
+          {sortedDocuments?.map((document) => (
+            <>
```
jadmnsaad: it's not good to add this fragment. A map function should return React nodes with the prop key. This is not really an option. What are we trying to do here?

jadmnsaad (changes requested):
array.map function should never return a fragment

jadmnsaad (approved):

PR #93, opened 2024-10-01
set up sort order for item types

djnunez-aot:
-setup sort order for item types to control the order of the submission package table


jadmnsaad (changes requested):
As per our conversation:
- We wanna sort in the schema
- We wanna find a better place to put the sort_order to ensure flexibility for each package type

jadmnsaad (approved):

jadmnsaad (approved):

PR #101, opened 2024-10-08
Add statuses to submission items

djnunez-aot:
-pass status to api from front end
-update the status of each item 
-added update_item function in the item service


submit-api/src/submit_api/services/item.py
```diff
@@ -11,3 +12,17 @@ def get_item_by_id(cls, item_id):
...
+        """Update item by id."""
+        session = db.session
+        item = cls.get_item_by_id(item_id)
+        if not item:
+            raise ValueError(f"Item with id {item_id} not found.")
+
+        for key, value in update_data.items():
+            setattr(item, key, value)
+
+        session.commit()
```
jadmnsaad: you can lose the session = db.session and the session.commit() and just do item.save() when you're done updating

submit-api/src/submit_api/services/submission/__init__.py
```diff
@@ -39,7 +40,14 @@ def create_submission(cls, item_id, request_data):
 
         submission_creator = cls.make_submission_creator(submission_type)
         submission_data = request_data.get("data")
-        return submission_creator.create(item_id, submission_data)
+        status = request_data.get("status")
+        # Update the item with the status from the request
+        if status is not None:
```
jadmnsaad: create a separate function for this added code and just call it in create_submission. This way it better adheres to the Single Responsibility Principle https://github.com/ryanmcdermott/clean-code-javascript?tab=readme-ov-file#functions-should-do-one-thing

submit-api/src/submit_api/services/submission/__init__.py
```diff
@@ -60,4 +68,10 @@ def edit_submission_form(cls, submission_id, request):
         submission = cls.get_submission_by_id_and_validate_edit(submission_id)
         submission.submitted_form.submission_json = request.get('data')
         submission.submitted_form.save()
+        item_id = request.get("item_id")
+        status = request.get("status")
+        # Update the item with the status from the request
+        if status is not None:
+            update_data = {"status": status}
+            ItemService.update_item(item_id, update_data)
```
jadmnsaad: not a good place to add this  code, this function is to validate not to do update operations

submit-web/src/components/SubmissionItem/ConsultationRecord/index.tsx
```diff
@@ -173,15 +184,17 @@ export const ConsultationRecord = () => {
     callSaveSubmission({
       data: {
         type: SUBMISSION_TYPE.FORM,
+        status: status || SUBMISSION_STATUS.COMPLETED.value,
```
jadmnsaad: I don't love that we are defaulting here, we shouldn't default at all, the status should be passed mandatorily

submit-web/src/components/SubmissionItem/ConsultationRecord/index.tsx
```diff
@@ -162,7 +166,14 @@ export const ConsultationRecord = () => {
     formState: { errors, dirtyFields },
   } = methods;
 
-  const saveSubmission = async (formData: ConsultationRecordForm) => {
+  const handleFormSubmit = (formData: ConsultationRecordForm) => {
```
jadmnsaad: handleCompleteForm

submit-web/src/components/SubmissionItem/ManagementPlanSubmission/index.tsx
```diff
@@ -156,6 +167,8 @@ export const ManagementPlanSubmission = () => {
     callSaveSubmission({
       data: {
         type: SUBMISSION_TYPE.FORM,
+        status: status || SUBMISSION_STATUS.COMPLETED.value,
```
jadmnsaad: we should never default in this case, we should never let the status be undefined

jadmnsaad (changes requested):
Added a few comments

submit-api/src/submit_api/services/item.py
```diff
@@ -11,3 +11,16 @@ def get_item_by_id(cls, item_id):
         """Get item by id."""
         item = ItemModel.find_by_id(item_id)
         return item
+
+    @classmethod
+    def update_item(cls, item_id, update_data):
+        """Update item by id."""
+        item = cls.get_item_by_id(item_id)
```
jadmnsaad: just for clarity item -> submission_item, because we are using the dict.items() method and devs might be confused

submit-api/src/submit_api/resources/submission.py
```diff
@@ -55,6 +54,10 @@ def post(submission_item_id):
         """Create a submission."""
         create_submission_data = CreateSubmissionRequestSchema().load(API.payload)
         created_submission = SubmissionService.create_submission(submission_item_id, create_submission_data)
+        item_id = create_submission_data.get("item_id")
+        status = create_submission_data.get("status")
+        if status:
+            SubmissionService.update_submission_item_status(item_id, status)
```
jadmnsaad, line 60: this should only be done at the service level. The resource only calls one service method

submit-api/src/submit_api/resources/submission.py
```diff
@@ -76,4 +79,8 @@ def patch(submission_id):
         """Edit a submission."""
         edit_submission_data = CreateSubmissionRequestSchema().load(API.payload)
         edited_submission = SubmissionService.edit_submission_form(submission_id, edit_submission_data)
+        item_id = edit_submission_data.get("item_id")
+        status = edit_submission_data.get("status")
+        if status:
```
jadmnsaad, line 84: same comment as above

submit-web/src/components/Submission/SubmissionStatusChip.tsx
```diff
@@ -47,7 +52,7 @@ export default function SubmissionStatusChip({
 }: {
   status: SubmissionStatus;
 }) {
-  const style = statusStyles[status];
+  const style = statusStyles[status] || statusStyles.NEW_SUBMISSION;
```
jadmnsaad, line 55: why are we defaulting like this? we shouldn't default to any status.

djnunez-aot: theres a error on mount wihout this

jadmnsaad (approved):
Approved with technical debt to improve our pattern particularly in the resources

PR #103, opened 2024-10-09
Document Upload UI Tweaks

djnunez-aot:
The text in this section was changed to “Drag and Drop a File Here or Select a File.”

Changed the comma at the end of the allowed files for a period. “Accepted file types: pdf, doc, docx, xlsx. Max. file size: 250 MB.”

TODO:
MP Upload section should allow only one document upload. The CR upload and supporting Doc Upload in the MP section can have more than one document. 

There is no file removal implemented yet so cannot replace, the previous management plan upload when a new one is uploaded

### DIFF
--- a/submit-web/src/components/FileUpload/index.tsx
+++ b/submit-web/src/components/FileUpload/index.tsx
@@ -9,14 +9,17 @@ export type FileUploadProps = {
   accept?: Accept;
   onDrop: (acceptedFiles: File[]) => void;
   error?: boolean;
+  maxSize?: number;
 };
 export const FileUpload = ({
   height = "10em",
   accept = {
-    pdf: [],
-    doc: [],
-    docx: [],
-    xls: [],
+    "application/pdf": [],
+    "application/msword": [],
+    "application/vnd.openxmlformats-officedocument.wordprocessingml.document":
+      [],
+    "application/vnd.ms-excel": [],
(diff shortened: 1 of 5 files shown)


submit-web/src/components/FileUpload/index.tsx
```diff
@@ -31,14 +31,14 @@ export const FileUpload = ({
...
-        Drag and Drop File(s) Here or{" "}
+        Drag and Drop a File Here or{" "}
         <span
           style={{
             textDecoration: "underline",
             color: "#1A5A96",
           }}
         >
-          Select File(s)
+          Select a Filea
```
jadmnsaad: typo

submit-web/src/components/SubmissionItem/ConsultationRecord/DocumentUploadSection.tsx
```diff
@@ -100,7 +100,7 @@ export const DocumentUploadSection = () => {
             color: EAOColors.ProponentDark,
           }}
         >
-          Accepted file types: pdf, doc, docx, xlsx, Max. file size: 250 MB.
+          Accepted file types: pdf, doc, docx, xlsx. Max. file size: 250 MB.
```
jadmnsaad, line 103: is this true, can we check that it can only accept these file types and not others.

And the max file size, is that also true? where is that set?

jadmnsaad (approved):

PR #104, opened 2024-10-09
Prevent form submission on enter

djnunez-aot:
-When user clicks enter prevent form submission

Not sure whats a more reusable way to do this maybe a form tag wrapper with this built in we import?

### DIFF
--- a/submit-web/src/components/Shared/Forms/common.tsx
+++ b/submit-web/src/components/Shared/Forms/common.tsx
@@ -5,7 +5,7 @@ type FormProps = {
   children: React.ReactNode;
 } & React.FormHTMLAttributes<HTMLFormElement>;
 
-export default function Form({ onSubmit, children, ...rest }: FormProps) {
+export default function Form({ children, ...rest }: FormProps) {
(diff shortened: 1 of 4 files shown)


jadmnsaad (approved):

jadmnsaad:
RE: Not sure whats a more reusable way to do this maybe a form tag wrapper with this built in we import?

That is one way, and react hook form has its own Form component, maybe that one doesn't have this problem?

submit-web/src/components/Shared/Forms/common.tsx
```diff
@@ -0,0 +1,23 @@
...
+    if (
+      event.key === "Enter" &&
+      (event.target as HTMLElement).tagName === "INPUT"
+    ) {
+      event.preventDefault();
+    }
+  };
+
+  return (
+    <form onSubmit={onSubmit} onKeyDown={handleKeyDown} {...rest}>
```
jadmnsaad: do u need to explicitly pass onSubmit? can't it be part of rest

djnunez-aot: good catch

PR #106, opened 2024-10-09
Only display status' for items are partially complete or completed

djnunez-aot:


submit-web/src/components/Submission/SubmissionItemTableRow.tsx
```diff
@@ -116,7 +117,19 @@ export default function SubmissionItemTableRow({
         <StyledTableCell align="right"></StyledTableCell>
         <StyledTableCell align="right"></StyledTableCell>
         <StyledTableCell align="right">
-          <SubmissionStatusChip status={status} />
+          <If
```
jadmnsaad: use Unless instead of If Then Else since your Then is empty

submit-web/src/components/Submission/SubmissionItemTableRow.tsx
```diff
@@ -116,7 +116,9 @@ export default function SubmissionItemTableRow({
         <StyledTableCell align="right"></StyledTableCell>
         <StyledTableCell align="right"></StyledTableCell>
         <StyledTableCell align="right">
-          <SubmissionStatusChip status={status} />
+          <When condition={SUBMISSION_STATUS.NEW_SUBMISSION.value != status}>
```
jadmnsaad: != -> !==

jadmnsaad: you could also use Unless here if u like

jadmnsaad (approved):

PR #108, opened 2024-10-15
Setup Submission Name, Status & Date Submitted Filters (UI only)

djnunez-aot:
-Setup the submission name filter

-Setup the status filter
-Setup dropdown with multiple statuses that can be selected

-Setup date submitted filter
-Request for date range picker but need MUI pro license to use it. Will discuss with MA 

-Added a zustand store to store all the filters

(integration with react query + zustand to connect to backend will be implemented once endpoints are created)


jadmnsaad (approved):
Well done, very clean

PR #110, opened 2024-10-16
Setup management plan search

djnunez-aot:
-setup management plan filter queries

-clean up ui to only search on enter


submit-web/src/hooks/api/useProjects.tsx
```diff
@@ -39,10 +39,26 @@ export const useLoadProjectsByProponentId = (proponentId?: number) => {
...
+  searchOptions?: Record<string, string | number>;
 };
-const getProjectsByAccount = ({ accountId }: GetProjectsByAccountParams) => {
+
+const getProjectsByAccount = ({
+  accountId,
+  searchOptions,
+}: GetProjectsByAccountParams) => {
+  // Initialize URL with base path and account ID
+  let url = `/projects/accounts/${accountId}`;
```
jadmnsaad: I don't think you have to manually pass the args to the url like this, I'm sure there's a params prop or something in axios with which u can pass these params  and axios will handle it

submit-api/src/submit_api/models/queries/project.py
```diff
@@ -13,24 +13,52 @@
...
         """Find projects by proponent_id"""
-        result = db.session.query(Project).filter(Project.proponent_id == proponent_id)
-        return result
+        query = db.session.query(Project).filter(
+            Project.proponent_id == proponent_id
+        )
+        return query.all()
+
+    @classmethod
+    def apply_search_filters(cls, query, search_options):
```
jadmnsaad: simplify to look like this
image

submit-api/src/submit_api/models/queries/project.py
```diff
@@ -13,24 +13,52 @@
...
+        if hasattr(search_options, 'status'):
+            query = cls._filter_by_status(query, search_options.status)
+        # Additional filters can be added here in the future
+        return query
+
+    @classmethod
+    def _filter_by_search_text(cls, query, search_text):
+        """Filter by search text across project name and package name."""
+        query = query.join(Package, AccountProject.packages).filter(
+            db.or_(
```
jadmnsaad: what's the usage of the or operation here?

djnunez-aot: im filtering through the submission names, (past & active submissions) like im trying to return the projects that have submissions which match the search_text the user inputted

jadmnsaad: sure, but why do u need the or?

- and: if condition A is true AND condition B is true return true
- or: if condition A is true OR condition B is true return true

why do u need it?

djnunez-aot: small over sight on my end just removed that

jadmnsaad (approved):

PR #115, opened 2024-10-21
Add submission status filtering - backend

djnunez-aot:
-filter by submission status enum


submit-api/src/submit_api/models/queries/project.py
```diff
@@ -47,15 +47,30 @@ def filter_by_search_criteria(cls, query, search_options):
...
         return query
+
+    @classmethod
+    def _filter_by_submission_status(cls, query, statuses):
+        """Filter by submission status using overlap."""
+        if statuses:
+            # Convert enum to string values
+            status_values = [status.value for status in statuses]
+            if status_values:
+                query = query.join(Package, AccountProject.packages).filter(
```
jadmnsaad: let's not join again on each filter, join is an expensive operation. in the main function join with Package there and in the filter functions just do filter

submit-api/src/submit_api/models/queries/project.py
```diff
@@ -47,15 +51,24 @@ def filter_by_search_criteria(cls, query, search_options):
         if not search_options:
             return query
```
jadmnsaad: maybe it's better to add the join with package at this spot

jadmnsaad (approved):

PR #123, opened 2024-10-28
Filter packages for each project with search options

djnunez-aot:
-For each project we should remove any packages that dont match the search criteria

-Added a function to manipulate the packages of each project in the query
-The filter is additive: if two statuses are selected, the results shows all MP with those two statuses only. Not all the statuses with one status or the other.

If I selected “New Submission” and “Submitted”, I should have no result. 

Can the Status label be aligned with the other labels and the input file made the same height, please? 

addressed this comment as well

### DIFF
--- a/submit-api/src/submit_api/models/queries/project.py
+++ b/submit-api/src/submit_api/models/queries/project.py
@@ -64,16 +64,15 @@ class ProjectQueries:
         ):
             return False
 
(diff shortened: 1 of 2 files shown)


jadmnsaad:
closed because this was addressed differently here https://github.com/bcgov/EPIC.submit/pull/124

PR #136, opened 2024-11-04
Changes for CRON and Email Queue Table added

saravanpa-aot:
Changes for CRON and Email queue template

**Changes in the Submit-API side**

- Added new email_queue table
- Added writing a pending record into the email_queue table

**Pending things in Submit-API Side**

- Convert and leverage constants for Pending, Template Name etc , Entity type etc.
- Use Enum in the table definitition
- remove CHES usage

Will do these changes in next PR

**Changes in the CRON side**

- Added new project , its build config , deployment config etc
- Added the code to process email

**Pending things in CRON Side**

- Leverage more constants
- Add CHES variables to deployment config 
- There is an issue with using the models from submit-api..used a decent workaround for now..fix this


submit-api/src/submit_api/services/package.py
```diff
@@ -161,15 +162,26 @@ def submit_package(cls, package_id):
             cls._update_items_status(package.items, ItemStatus.SUBMITTED.value, session)
             cls._update_package_status(package_id, session, package)
             cls._update_package_submission_details(package, session)
+            cls._create_email_queue_record(package, session)
             session.flush()
             session.commit()
+        # TODO: Remove this once we have a generic email service
         cls.send_package_submission_email_confirmation(package)
```
jadmnsaad, line 169: so we have to remove this then

djnunez-aot (approved):

PR #180, opened 2024-11-27
Integrate review component with api

jadmnsaad:
- Add useSubmissionReview hook
- Add SubmissionReview model
- Add db migration

left to be done: setting default values


submit-api/src/submit_api/services/item.py
```diff
@@ -54,16 +54,23 @@ def update_submission_item(cls, item_id, update_data):
     def get_or_create_active_item_review(cls, item_id) -> SubmissionReview:
         """Get item by id."""
         _ = cls.get_item_by_id(item_id)
-        review = SubmissionReview.get_by_item_id(item_id)
+        review = SubmissionReview.get_active_review_by_item_id(item_id)
+        print(review)
```
djnunez-aot: print statements here the rest looks good

djnunez-aot (approved):

submit-api/src/submit_api/services/item.py
jadmnsaad: Thanks for the catch

PR #183, opened 2024-11-28
Notes api

djnunez-aot:
-Create notes api


submit-api/src/submit_api/models/note.py
```diff
@@ -0,0 +1,24 @@
...
+class Note(BaseModel):
+    """Definition of the submitted documents entity."""
+
+    __tablename__ = "internal_staff_documents"
+
+    id = Column(db.Integer, primary_key=True, autoincrement=True)
+    note = Column(db.String(255), nullable=False)
+    item_id = Column(db.Integer, db.ForeignKey("items.id"), nullable=False)
+    created_by = Column(db.String, db.ForeignKey("users.auth_guid"), nullable=True)
+    created_date = Column(db.DateTime, nullable=False)
```
jadmnsaad: I think created_date is already inherited from BaseModel so you don't need to add it again

submit-api/src/submit_api/resources/staff/note.py
```diff
@@ -0,0 +1,57 @@
...
+    @API.response(
+        code=HTTPStatus.CREATED, model=note, description="Created a staff note"
+    )
+    @API.response(HTTPStatus.NOT_FOUND, "Not found")
+    @auth.require
+    @cors.crossdomain(origin="*")
+    def post(submission_item_id):
+        """Create a staff note."""
+        create_note_data = PostNote().load(API.payload)
+        created_note = (NoteService.create_note(submission_item_id, create_note_data))
```
jadmnsaad: why the  ()?, I think this has the effect of creating a tuple. Do you need that?

submit-api/src/submit_api/services/note.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+"""Service for internal staff document service."""
+from submit_api.exceptions import ResourceNotFoundError
+from submit_api.models import Item
+from submit_api.models.note import Note as NoteModel
+from submit_api.utils.token_info import TokenInfo
+
+
+class NoteService:
```
jadmnsaad: In theory, we could have notes for any entity. That's why I would change any naming from "Notes" -> "SubmissionNote" or "SubmissionItemNote". And that goes everywhere (the resource, the service, the model, the schema etc.) Just so we don't get caught in an awkward position when we have to add a Note  to some other entity like Package per se.

submit-api/src/submit_api/resources/staff/note.py
djnunez-aot: no we dont

jadmnsaad (approved):

PR #207, opened 2024-12-20
setup front end for starting MP review & CR check

djnunez-aot:
-add modal confirmation when beginning MP review or CR check

-add review_start_date


submit-web/src/components/Submission/DocumentRow.tsx
```diff
@@ -4,19 +4,39 @@ import { useState } from "react";
...
     submitted_document: { name, url },
     version,
     submitted_by,
   } = documentSubmission;
 
+  const openVerificationModal = () => {
+    //todo: setup onConfirm
+    openModal(
+      <ConfirmationModal
+        onConfirm={() => {}}
```
jadmsaadaot, line 31: okay, so what happens when they click confirm?

They should proceed to open the document right?

djnunez-aot, line 31: update status then open yes

PR #210, opened 2024-12-27
Added invitation token creation

saravanpa-aot:


jadmnsaad (dismissed):

submit-api/src/submit_api/models/invitations.py
```diff
@@ -0,0 +1,69 @@
...
+from .db import db
+
+
+class Invitations(BaseModel):
+    """Definition of the Invitation Token entity."""
+
+    __tablename__ = 'invitations'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    account_id = Column(Integer, ForeignKey('accounts.id'), nullable=False)
```
jadmnsaad, line 22: so Invitation can only be created for existing accounts. Which is all good but just making sure that's the decision you're making.

Right now we create the account in the registration process. This change will require us to have a user management step where somebody needs to know the proponent_id then they need to create the account and assign projects then they need to create the token. The user registration step then registers the account_user only perhaps.

saravanpa-aot: Good point , lets discuss this over a call..
I wasnt very sure of how to tackle this , so worth a discussion

PR #213, opened 2024-12-30
Update MP & CR Plans on starting review

djnunez-aot:
Start review_start_date on mp review confirmation
start cc_start_date on cc_check confirmation
-Setup methods for mp_review_start and cc_check_start
-Update front end accordingly


submit-web/src/components/Submission/DocumentRow.tsx
```diff
@@ -17,27 +21,45 @@ export default function DocumentRow({
...
+      },
+    });
 
   const openVerificationModal = () => {
     //todo: setup onConfirm
     setOpenModal(
       <ConfirmationModal
         onConfirm={() => {
           setCloseModal();
+          updateStateSubmissionPackage({
```
jadmnsaad, line 53: currently with this:

when the modal pops up
- user clicks confirm
- modal immediately closes
- an api call happens in the backend

this way there's not error handling if the api call fails, and the user doesn't see anything loading

what we probably need:
- user clicks confirm
- confirm button loads while api call is pending
- after success the modal is closed

submit-web/src/components/Submission/DocumentRow.tsx
```diff
@@ -17,27 +21,45 @@ export default function DocumentRow({
...
+    name === SUBMISSION_ITEM_TYPE.CONSULTATION_RECORD;
+
+  const { mutate: updateStateSubmissionPackage } =
+    useUpdateStateSubmissionPackage({
+      onError: () => {
+        notify.error("Failed to update management plan");
+      },
+    });
 
   const openVerificationModal = () => {
```
jadmnsaad: it's a confirmation modal not a verification modal

PR #266, opened 2025-01-28
Add submit web link in update request email

jadmnsaad:
- update charts: deployment config, configmap and values to add the web url variable
- get and pass the web url variable to the email renderer
- Update email template to include link to submit web


saravanpa-aot (commented):
does the BASE_APP_URL in the config.py for the same purpose?

jadmnsaad:
> does the BASE_APP_URL in the config.py for the same purpose?

yeah I think BASE_APP_URL is equivalent to WEB_URL that I just added to submit-cron. I didn't find a BAS_APP_URL in submit-cron that's in submit-api. Am I missing something?

PR #277, opened 2025-02-04
Added activity logging

saravanpa-aot:
for https://eao-dst.atlassian.net/browse/SUBMIT-264

- Added the model and migration scripts
- API for getting the activity
- Added one activity for submission creation


submit-api/src/submit_api/services/activity_log_service.py
```diff
@@ -0,0 +1,84 @@
...
+        - `entity_type`: The type of entity (e.g., 'submission', 'file_upload').
+        - `entity_id`: The specific entity ID.
+        - `for_staff`: If False, only public logs are returned.
+
+        Returns a list of logs.
+        """
+        query = ActivityLog.query.filter_by(entity_type=entity_type, entity_id=entity_id)
+
+        if not for_staff:
+            query = query.filter(ActivityLog.visibility == "public")
```
jadmnsaad, line 65: magic string "public"

submit-api/src/submit_api/services/package.py
```diff
@@ -304,8 +306,21 @@ def submit_package(cls, package_id):
         with session_scope() as session:
             package = cls._get_and_validate_complete_package(package_id)
             if package.submitted_on:
-                return cls._resubmit_package(package, session)
-            return cls._submit_package(package, session)
+                submitted_package: PackageModel = cls._resubmit_package(package, session)
+            else:
+                submitted_package: PackageModel = cls._submit_package(package, session)
+
+            ActivityLogService.log_activity(
```
jadmnsaad, line 313: better this be in a separate method to avoid cluttering submit_package. Maybe we can create def _submit_package_activity_log

submit-api/src/submit_api/enums/activity_type.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class ActivityTypeEnum(enum.Enum):
+    """Enum for activity type statuses."""
+
+    SUBMISSION = 'SUBMISSION'
+    USER = 'USER'
+
+
+class ActorTypeEnum(enum.Enum):
```
jadmnsaad, line 17: Does this mean staff or entity user? If so then the Staff here is also a user, maybe we can change to something like:
ENTITY_USER
STAFF_USER

PR #283, opened 2025-02-06
More activity logs added

saravanpa-aot:
Added activity logs for most of the actions

### DIFF
--- a/submit-api/src/submit_api/services/submission_review.py
+++ b/submit-api/src/submit_api/services/submission_review.py
@@ -171,8 +171,6 @@ class SubmissionReviewService:
         approval_processor(item, session)
         cls._update_package_status(item.package_id, session)
         cls._update_item_submissions_status(SubmissionStatus.APPROVED, session, item=item)
-        if item.type.name == SubmissionItemType.MANAGEMENT_PLAN_FORM.value:
-            cls._log_activity_mp_review(item, session, approved=True)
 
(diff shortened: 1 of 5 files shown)


submit-api/src/submit_api/services/submission_review.py
```diff
@@ -169,6 +171,9 @@ def approve_submission(cls, item_id, session):
         approval_processor(item, session)
         cls._update_package_status(item.package_id, session)
         cls._update_item_submissions_status(SubmissionStatus.APPROVED, session, item=item)
+        if item.type.name == SubmissionItemType.MANAGEMENT_PLAN_FORM.value:
```
saravanpa-aot: I wasnt sure if the approval happens just for management plan..so just added a check ;might not be necessary..plz lmk

jadmnsaad: 1- we haven't implemented approve Management plan yet
2- that pattern here is an type-processor pattern so we don't have to add if statements. look into cls._get_submission_item_approval_processor(item) and you will see all the item types that can be approved at the moment.

jadmnsaad: image
This will have to he extended and we should add approve_management_plan here. And in approve_management_plan we should add that log. We should not add anything specific in approve_submission

jadmnsaad (approved):

PR #298, opened 2025-02-18
Add submission sub table in items table

jadmnsaad:
- Add submission versions subtable
- Delete submissions when they are replaced and pending


submit-api/src/submit_api/models/submission.py
```diff
@@ -46,6 +46,7 @@ class Submission(BaseModel):
     major_version = Column(db.Integer, nullable=False, default=1)
     minor_version = Column(db.Integer, nullable=False, default=1)
     active = Column(db.Boolean, nullable=False, default=True)
+    deleted = Column(db.Boolean, nullable=False, default=False)
```
saravanpa-aot, line 49: is_deleted might be better..but not too sure if there is a standard
But do you think if there is a scope for including this in the status somehow? like there are two flags active,deleted and status which feels like does the same thing and over time we might forget why we went for each of this separately..

jadmnsaad: Sure adding it in the status or _deleted work. Do you have a strong feeling about this and want to do that or are these just for consideration?

 I thought I wanted to separate the deleted flag from the normal business logic, which includes status.

saravanpa-aot (approved):

PR #301, opened 2025-02-20
Changes to accept update for update request

VineetBala-AOT:
Ticket: https://eao-dst.atlassian.net/browse/SUBMIT-320

- Introduced a new "ACCEPTED" status for update requests.
- Added a button for staff to accept submission requests.
- Enabled entities to modify update requests unless they have been accepted by staff.


submit-web/src/components/Submission/SubmissionItemTableRow/ProponentSubmissionItemTableRow.tsx
```diff
@@ -133,7 +133,7 @@ export default function ProponentSubmissionItemTableRow({
               submissionPackage?.submitted_on &&
               submissionPackage.update_requests.filter(
                 (updateRequest) =>
-                  updateRequest.status === UPDATE_REQUEST_STATUS.OPEN.value,
```
jadmsaadaot, line 136: I think entities want to only see Open requests

submit-api/src/submit_api/services/package.py
```diff
@@ -342,6 +342,23 @@ def _resubmit_package(cls, package, session):
         cls._log_activity_submission(package, ActivityActionType.UPDATED_SUBMISSION.value, session)
         return package
 
+    @classmethod
+    def accept_update_request(cls, package_id):
```
jadmnsaad, line 346: have you thought about updating the active attribute to false when we accept an udpate request

submit-web/src/components/Submission/SubmissionItemTableRow/ProponentSubmissionItemTableRow.tsx
VineetBala-AOT, line 136: Thanks. The ticket said that entity should be able to submit unless the update request is accepted so I made this change so that they can edit the data unless the request is accepted by staff.

submit-api/src/submit_api/services/package.py
VineetBala-AOT: Sure let me check on that.

submit-api/src/submit_api/services/package.py
```diff
@@ -342,6 +342,23 @@ def _resubmit_package(cls, package, session):
...
+            package = cls._get_and_validate_complete_package(package_id)
+
+            current_app.logger.info(f"Accepting update reuqest for package {package.id}")
+            open_update_requests = [
+                request for request in package.update_requests
+                if request.status in {UpdateRequestStatus.OPEN.value,
+                                      UpdateRequestStatus.PENDING_REVIEW.value}
+            ]
+            if not open_update_requests:
+                raise BadRequestError("Cannot accept a package that has no open or pending update requests")
```
jadmnsaad: "Cannot accept an update request* that is not pending"

submit-api/src/submit_api/services/package.py
```diff
@@ -342,6 +342,23 @@ def _resubmit_package(cls, package, session):
...
+    @classmethod
+    def accept_update_request(cls, package_id):
+        """Submit the package by updating its status and items."""
+        with session_scope() as session:
+            package = cls._get_and_validate_complete_package(package_id)
+
+            current_app.logger.info(f"Accepting update reuqest for package {package.id}")
+            open_update_requests = [
+                request for request in package.update_requests
+                if request.status in {UpdateRequestStatus.OPEN.value,
```
jadmnsaad: Do we wanna accept update requests that are not even pending?

When EAO create an update request the status is OPEN
When Entity resubmits package the status becomes PENDING

I think it makes sense to only accept pending requests

submit-api/src/submit_api/models/update_request.py
```diff
@@ -25,6 +25,7 @@ class UpdateRequestStatus(enum.Enum):
     OPEN = 'OPEN'
```
jadmnsaad, line 25: I'm thinking now or later maybe we can change this status to something like "CREATED" which is probably more accurate for our use case.

jadmnsaad: not a blocker for merge, this could definitely be done in a separate task

submit-api/src/submit_api/services/package.py
VineetBala-AOT: Updated to set the attribute to false

submit-api/src/submit_api/services/package.py
VineetBala-AOT: Updated, thanks

submit-api/src/submit_api/services/package.py
VineetBala-AOT: Updated

jadmnsaad (approved):

PR #309, opened 2025-02-25
Update to Submission item table

djnunez-aot:
Make the Form/document name larger

Left-align the Uploaded by label and column content

Left-align the Actions label and the actions, leaving 16px gap between the text and the right edge of the table

Move all the columns but the first to the right so they are only as large as required and close together.

-Make sure its responsive


submit-web/src/components/Submission/DocumentRow/index.tsx
```diff
@@ -55,7 +55,7 @@ export default function DocumentRow({
   return (
     <>
       <TableRow sx={{ "& > *": { borderBottom: "unset" } }}>
-        <SubmitTableCell>
+        <SubmitTableCell colSpan={10}>
```
jadmnsaad: the colSpan usage in these changes is not the right way. there are for sure not 10 columns so why colSpan 10?

jadmnsaad (changes requested):
wrong usage of colSpan

jadmnsaad (approved):

PR #322, opened 2025-03-05
Setup multiple select in new user form

djnunez-aot:


jadmnsaad (dismissed):

submit-web/src/components/UserManagement/NewUser/NewUserForm.tsx
```diff
@@ -134,6 +155,21 @@ export default function NewUserForm() {
               <ControlledRadioGroup name="applicationAccess">
                 <FormOptions error={Boolean(errors["applicationAccess"])} />
               </ControlledRadioGroup>
+
+              <When condition={selectedRole === "CollaboratorSpecific"}>
+                <ControlledMultiSelect
+                  multiple
+                  selectAll
+                  name="project_ids"
+                  options={
```
jadmnsaad: you can extract this out into a const and use useMemo so it doesn't recalculate on every render

jadmnsaad (approved):

PR #329, opened 2025-03-07
Integrate invitation api with new user page

djnunez-aot:
Setup the integration for the invitation api with then new user page


submit-web/src/components/UserManagement/NewUser/FormOptions.tsx
```diff
@@ -7,37 +7,42 @@ type FormOptionsProps = {
   disabled?: boolean;
 };
 
+export enum Role {
```
jadmnsaad: this seems useful across the app, can we take it to constant file, somewhere shared and common

jadmnsaad (dismissed):

jadmnsaad (approved):

PR #332, opened 2025-03-10
Setup user table updates

djnunez-aot:
Update typing for user chip
Update data to display status and role correctly


jadmnsaad (dismissed):

submit-web/src/components/UserManagement/entity/UserTableRow.tsx
```diff
@@ -14,10 +15,10 @@ export default function UserTableRow({ user }: { user: AccountUser }) {
         <SubmitLink>{user.full_name}</SubmitLink>
       </PlainTableCell>
       <PlainTableCell align="left" width={"25%"}>
-        {user.position}
+        {user.roles[0] && roleDetails[user.roles[0].role_name].label}
```
jadmnsaad, line 18: how come we chose only to show the role at index 0?

djnunez-aot: we could show all roles but from my understanding they only have one role, I was confused on why it was an array

jadmnsaad (approved):

PR #366, opened 2025-03-18
Make generate url textfield the same as the others

jadmnsaad:
- remove variant standard
- make name normal and not a link

covering second ticket:
- Make package submission success message appear on every submit

### DIFF
--- a/submit-api/src/submit_api/services/invitation_service.py
+++ b/submit-api/src/submit_api/services/invitation_service.py
@@ -46,7 +46,7 @@ class InvitationService:
             session.flush()
             invitation = InvitationService._create_invitation_record(invite_data,
                                                                      role,
-                                                                     account.id,
+                                                                     account,
                                                                      token,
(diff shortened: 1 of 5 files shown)


submit-api/src/submit_api/services/invitation_service.py
```diff
@@ -46,7 +46,7 @@ def create_invitation(invite_data):
             session.flush()
             invitation = InvitationService._create_invitation_record(invite_data,
                                                                      role,
-                                                                     account.id,
+                                                                     account,
```
saravanpa-aot, line 49: dont u have to change the method to accept account instead of account id

jadmnsaad: yes, because in that service I need to check if account has users to decide the is_first_time flag. And If I pass that separately to the method I get a too many arguments linting error

saravanpa-aot (approved):

PR #367, opened 2025-03-18
changes for fixing no project name for invitations with package id

saravanpa-aot:
When collaborators are added to project , there is no project id..fetched project name using package id as a part of this change.


VineetBala-AOT (approved):

submit-cron/src/submit_cron/services/invitation_email_service.py
```diff
@@ -15,7 +17,14 @@ class InvitationEmailService:  # pylint: disable=too-few-public-methods
     @classmethod
     def prepare_invitation_email_notification(cls, invitation: InvitationsModel) -> EmailDetails:
         """Prepare email details for update request creation."""
-        project_name = cls.get_project_names(invitation.project_ids)
+        invitation_action_text = "join"   # decides the content of the email.
```
jadmsaadaot, line 20: magic string

jadmnsaad (approved):

PR #384, opened 2025-03-20
Fix duplicate submission name displayed while ading new user

VineetBala-AOT:
Ticket: https://eao-dst.atlassian.net/browse/SUBMIT-440

- Updated the logic to ensure that package names and IDs are retrieved based on the latest version of the original package.
- This change prevents duplicate package names from appearing when adding a new user for a specific submission.

### DIFF
--- a/submit-api/src/submit_api/models/queries/package.py
+++ b/submit-api/src/submit_api/models/queries/package.py
@@ -17,6 +17,7 @@ from submit_api.enums.item_status import ItemStatus
 from submit_api.models import AccountProject, db
 from submit_api.models.package import Package as PackageModel
 from submit_api.models.package import PackageStatus
+from submit_api.models.package_version import PackageVersion
 
 
 # pylint: disable=too-few-public-methods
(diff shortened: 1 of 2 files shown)


submit-api/src/submit_api/models/queries/package.py
```diff
@@ -132,7 +133,21 @@ def update_package_status(package_id, session, package=None):
 
     @classmethod
     def get_account_project_packages(cls, account_id: int):
```
jadmnsaad: what will happen to the components that were using this method with the old query

VineetBala-AOT: I checked, and this method is only used in the New User Form and Update Role Form. The corresponding hook is useGetAccountPackagesByAccountId.

jadmnsaad: okay very well, could we plz change the name of this method to get_account_project_unique_packages

jadmnsaad: or %latest_packages

VineetBala-AOT: Changed the name of the method, thanks.

jadmnsaad (approved):

PR #388, opened 2025-03-25
Display Project name on proponent registration

djnunez-aot:
-Use proponent id to display project name on sign up


submit-api/src/submit_api/services/invitation_service.py
```diff
@@ -206,6 +206,14 @@ def get_valid_invitation(token):
         if invitation.expiry_date < datetime.datetime.utcnow():
             return {"error": "Invitation has expired"}, False
 
+        # Retrieve account from account_id
+        account = AccountModel.query.filter_by(id=invitation.account_id).first()
```
jadmnsaad: Isn't the account already joined to the invitation
image

jadmnsaad: can't we in the schema use fields.Pluck on the account to get the proponent_id? or a even a method to just access it and return it

djnunez-aot: i used fields.pluck and it was not returning the proponent_id

jadmnsaad (dismissed):

jadmnsaad (approved):

PR #393, opened 2025-03-26
Create and apply backend authorization checks for proponent users

jadmnsaad:
- Add has one or more permission decorator
- Add has permission service
- Add is assigned to package service


saravanpa-aot (commented):
looks good so far..lmk when its a full PR..will have a detailed look..

saravanpa-aot (approved):

PR #412, opened 2025-03-31
Display Previously Failed status for CR with failed submissions in history

jadmnsaad:
- Add GET failed submissions by item id
- Add Previously Failed status
- Add custom status cell for CR items
- Call endpoint and display status accordingly

### DIFF
--- a/submit-api/src/submit_api/resources/staff/submitted_document.py
+++ b/submit-api/src/submit_api/resources/staff/submitted_document.py
@@ -82,6 +82,5 @@ class ItemFailedDocuments(Resource):
     @cors.crossdomain(origin="*")
     def get(item_id):
         """Get all failed documents by item id."""
-
         documents = DocumentService.get_failed_documents_by_item_id(item_id)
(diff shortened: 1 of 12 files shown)


djnunez-aot (approved):

submit-api/src/submit_api/resources/staff/submitted_document.py
```diff
@@ -62,3 +62,25 @@ def get():
 
         documents = DocumentService.get_all_documents(search_options)
         return SubmittedDocumentByProjectSchema(many=True).dump(documents), HTTPStatus.OK
+
+
+@cors_preflight("GET, OPTIONS")
```
saravanpa-aot, line 67: i would probably make a generic api with a status filter here unless failed submission are totally a different deal
like item/:item_id?status=failed

jadmnsaad: I had a similar thought. But I thought this endpoint is akin to check if this item has failed in the past. And I just needed that and didn't need a generic endpoint. I thought we could always refactor if we needed.

PR #421, opened 2025-04-02
Fix Breadcrumbs

djnunez-aot:
Update breadcrumbs to

User Management > [user name]

User Management > Add New User


jadmnsaad (changes requested):
I don't think we need to do the layout pattern here. Keeping them separate might be lighter

jadmnsaad (approved):

PR #455, opened 2025-04-14
Do not update status for Contact information when it is submitted

djnunez-aot:
-Remove the status being passed down when contact information is updated (only after submission)


submit-web/src/components/SubmissionItem/ContactInformation/ContactInformationEntityView.tsx
```diff
@@ -115,7 +123,7 @@ export const ContactInformationEntityView = () => {
     const request = {
       type: SUBMISSION_TYPE.FORM,
       data: formData,
-      status: SUBMISSION_ITEM_STATUS.COMPLETED.value,
+      status: isSubmitted ? "" : SUBMISSION_ITEM_STATUS.COMPLETED.value,
```
jadmnsaad: I think perhaps passing undefined is better than passing ""

jadmnsaad (approved):

PR #486, opened 2025-04-28
Fix Revision Required Status filter

djnunez-aot:


submit-api/src/submit_api/models/queries/account_project.py
```diff
@@ -184,8 +193,41 @@ def _filter_by_search_text(cls, query, search_text):
...
+    def _revision_required_filter(cls):
+        """Builds the filter for packages requiring revision."""
+        return and_(
+            ~Package.status.op("@>")([
+                PackageStatus.COMPLETED.value,
+                PackageStatus.PARTIALLY_COMPLETED.value,
+            ]),
+            exists().where(
+                and_(
+                    UpdateRequest.submission_package_id == Package.id,
```
jadmnsaad: let's make sure we join with UpdateRequest first

jadmnsaad (changes requested):
Join with update request

submit-api/src/submit_api/models/queries/account_project.py
```diff
@@ -184,8 +192,39 @@ def _filter_by_search_text(cls, query, search_text):
...
+    def _revision_required_filter(cls, query):
+        """Joins UpdateRequest and filters for packages requiring revision."""
+        return query.filter(
+            ~Package.status.op("@>")([
+                PackageStatus.COMPLETED.value,
+                PackageStatus.PARTIALLY_COMPLETED.value,
+            ])
+        ).join(
+            UpdateRequest,
+            and_(
```
jadmnsaad: I don't think this and should be part of the join operation. It probably should be in a filter after the join

jadmnsaad (approved):

PR #493, opened 2025-04-30
Continue to form preview & document preview on start later

djnunez-aot:
-If Eao selects start later it will allow the user preview the document or view the form without starting an official review


submit-web/src/components/Submission/SubmissionItemReviewConfirmation.tsx
```diff
@@ -85,11 +85,14 @@ export default function SubmissionItemReviewConfirmation({
             },
           });
         }}
+        onCancel={() => {
```
jadmnsaad: and what if we wanted cancel to just cancel on do nothing? With this change cancel will always click onClick

jadmnsaad (approved):

PR #507, opened 2025-05-06
fix restrict upload to only one management plan

VineetBala-AOT:
Ticket: https://eao-dst.atlassian.net/browse/SUBMIT-528

- Implemented a check to prevent uploading multiple management plans. If a management plan is already uploaded, further uploads are blocked, and an error notification is displayed.

### DIFF
--- a/submit-web/src/components/SubmissionItem/ManagementPlanSubmission/ManagementPlanProponentView/DocumentUploadSection.tsx
+++ b/submit-web/src/components/SubmissionItem/ManagementPlanSubmission/ManagementPlanProponentView/DocumentUploadSection.tsx
@@ -53,17 +53,6 @@ export const DocumentUploadSection = () => {
   }, [submissionItem, getDocumentSubmissions, initializeFiles]);
 
   const handleOnDrop = (acceptedFiles: File[], folder: string) => {
-    const isManagementPlanFolder =
-      folder === MANAGEMENT_PLAN_DOCUMENT_FOLDERS.MANAGEMENT_PLAN;
-
-    if (
-      isManagementPlanFolder &&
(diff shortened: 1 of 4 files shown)


submit-web/src/components/SubmissionItem/ManagementPlanSubmission/ManagementPlanProponentView/DocumentUploadSection.tsx
```diff
@@ -53,6 +53,17 @@ export const DocumentUploadSection = () => {
   }, [submissionItem, getDocumentSubmissions, initializeFiles]);
 
   const handleOnDrop = (acceptedFiles: File[], folder: string) => {
+    const isManagementPlanFolder =
```
jadmnsaad: when we add a file we do two things:
1- we call addPendingFile which updates the file store
2- we call setValue on the schema.
image

These two should behave together. Find a way to pass a prop of max length to the FileUpload so the on Drop and the setValue are not called from the start

jadmnsaad: find a way to utilize this
image

jadmnsaad (changes requested):
Use maxFiles on the DropZone instead of doing this

submit-web/src/components/SubmissionItem/ManagementPlanSubmission/ManagementPlanProponentView/DocumentUploadSection.tsx
VineetBala-AOT: Refactored code to pass and utilize maxFiles

jadmnsaad (dismissed):

jadmnsaad (approved):

PR #508, opened 2025-05-06
Setup Loaders For Plan Details

djnunez-aot:
-When loading the account projects we should display a loader to prevent having the user create a new plan by accident

### DIFF
--- a/submit-web/src/components/NewManagementPlan/PlanDetails.tsx
+++ b/submit-web/src/components/NewManagementPlan/PlanDetails.tsx
@@ -29,34 +29,29 @@ export const PlanDetails = ({ onSubmit }: PlanDetailsProps) => {
         formData?.main_condition?.condition_number
     );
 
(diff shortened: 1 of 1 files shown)


submit-web/src/components/NewManagementPlan/PlanDetails.tsx
```diff
@@ -30,15 +31,32 @@ export const PlanDetails = ({ onSubmit }: PlanDetailsProps) => {
...
+      <If condition={isLoading}>
+        <Then>
+          <Grid
+            container
+            sx={{
+              padding: "16px 0px",
+            }}
+            spacing={3}
+          >
+            <Grid item xs={12}>
```
jadmnsaad: early returns are easier to read.
image

jadmnsaad (changes requested):
use an early return

jadmnsaad (approved):

PR #512, opened 2025-05-08
SUBMIT-527 Documents displaying to the EAO before the Holder re-submits

djnunez-aot:
https://eao-dst.atlassian.net/jira/software/c/projects/SUBMIT/boards/3?selectedIssue=SUBMIT-527)](https://eao-dst.atlassian.net/browse/SUBMIT-527

Add submitted_submissions property for the eao side. This property will display all the most recent submitted submissions(active or inactive) for the eao.

This is to prevent displaying pending submissions from the proponents most recently updated(but not submitted) info.


submit-api/src/submit_api/services/package.py
```diff
@@ -272,8 +272,12 @@ def update_submission_status(package, status, session):
             raise BadRequestError("Invalid status")
         submissions = [submission for item in package.items for submission in item.submissions]
         for submission in submissions:
-            submission.status = status
-            session.add(submission)
+            if submission.status == SubmissionStatus.PENDING_REPLACEMENT:
+                submission.active = False
+                submission.status = status
```
jadmnsaad: this should probably be "Replaced" no?

djnunez-aot: we can add that when its set to active false its gone entirely except for previous versions but yes we can add that in

submit-api/src/submit_api/schemas/item.py
```diff
@@ -147,8 +166,35 @@ class Meta:  # pylint: disable=too-few-public-methods
 
         unknown = EXCLUDE
 
-    internal_staff_documents = fields.Nested(InternalStaffDocumentSchema,
+    internal_staff_documents = fields.Nested(InternalStaffDocumentSchema, 
                                              data_key="internal_staff_documents", many=True)
     review = fields.Nested(SubmissionReviewSchema, data_key="review")
     notes = fields.Nested(SubmissionItemNote, data_key="notes", many=True)
     review_start_date = fields.DateTime(data_key="review_start_date")
+    submitted_submissions = fields.Method('get_submitted_submissions')
```
jadmnsaad: it should be called submissions when sent to the UI

submit-web/src/components/SubmissionItem/ManagementPlanSubmission/ManagementPlanStaffView/index.tsx
```diff
@@ -25,8 +25,8 @@ export const ManagementPlanSubmissionStaffView = () => {
     itemId: Number(submissionItemId),
   });
 
-  const formSubmission = submissionItem?.submissions?.find(
-    (submission) => submission.type === SUBMISSION_TYPE.FORM,
+  const formSubmission = submissionItem?.submitted_submissions?.find(
```
jadmnsaad: why? why not send it back as submissions and keep it as submissions here

jadmnsaad (dismissed):

PR #518, opened 2025-05-12
Tasks SUBMIT-532 , SUBMIT-534 Changes for new Email

saravanpa-aot:
https://eao-dst.atlassian.net/browse/SUBMIT-534
https://eao-dst.atlassian.net/browse/SUBMIT-532

1. Added new email template
2. Removed constants which were duplicated in submit-api and in submit-cron..Now reusing the submit-api in submit-cron
3. Changes for the new email
4. fixed the issue with marshmallow versioning


jadmnsaad (dismissed):

submit-cron/templates/management_plan_submission_notify_staff.html
```diff
@@ -0,0 +1,79 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>New Management Plan Submission</title>
```
jadmnsaad: Are we still using the word Management Plan for this or are we generalizing it

saravanpa-aot: good find..but i think its the jinja html title which might not be seen anywhere.. i will change it anyways

PR #541, opened 2025-05-26
SUBMIT-537 Setup Conditional Box

djnunez-aot:
[Description](https://eao-dst.atlassian.net/jira/software/c/projects/SUBMIT/boards/3?selectedIssue=SUBMIT-537)

Setup Grey box


submit-web/src/routes/proponent/_proponentLayout/projects/$projectId/_projectLayout/submission-packages/$submissionPackageId/_submissionLayout/index.tsx
```diff
@@ -235,13 +253,44 @@ export default function SubmissionPage() {
               >
                 <ItemsTable submissionPackage={submissionPackage} />
               </Box>
-              <When condition={isPackageSubmitted && openRequests.length === 0}>
+              <When
```
jadmnsaad: Seems to me that these two boxes shouldn't appear together ever right? So it would probably be better to a Switch/Case logic to guarantee that only one will appear.

jadmnsaad (dismissed):

saravanpa-aot (commented):
is there a better way to handle this string in a common place?

jadmnsaad (approved):

PR #542, opened 2025-05-26
SUBMIT-473 User email check

djnunez-aot:
When a Project Admin clicks the + Add New User button, adds a user, and clicks Save, we must check all the existing users' email addresses to verify whether the same email is already associated with a user.

If the newly added user email address is already in use and the user is active, show a modal with the following text:

This email address already has an active user in your EPIC.submit account. 

To edit a user's access permissions:

    Navigate to the User Management table

    Find the user by name

    Click 'View/Edit User Access' to open their details page

    Select 'Edit Access' to modify permissions or manage submission collaborators

Button: Close

If the newly added user email address is already in use and the user is invited, show a modal with the following text:

This email address already has a pending invitation to EPIC.submit.

To resend the invitation:

    Go to the User Management table

    Locate the user by their name

    Click the 'Resend Email Invite' button

Once sent, the user will receive a new invitation email with instructions to join EPIC.submit.

Button: Close

If the newly added user email address is not already in use, send the email with the magic link as in the current flow.

### DIFF
--- a/submit-api/src/submit_api/services/invitation_service.py
+++ b/submit-api/src/submit_api/services/invitation_service.py
@@ -2,6 +2,7 @@
 import datetime
 import uuid
 from urllib.parse import urljoin
+from http import HTTPStatus
 
 from flask import current_app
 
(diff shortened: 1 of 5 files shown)


saravanpa-aot (commented):
I think this is more if an API validation check ;doing in UI is okay ;but IMHO we will have to do the same in API as well...

jadmnsaad:
LGTM

submit-api/src/submit_api/services/invitation_service.py
```diff
@@ -30,20 +29,49 @@ def generate_uuid_token():
         """Generate a unique UUIDv4 token."""
         return str(uuid.uuid4())
 
+    @staticmethod
+    def _check_existing_user(email, account_id):
+        """Check if a user with the given email already exists in the account."""
+        # Check for active users
+        existing_user = AccountUserService.get_users_by_account(
```
jadmnsaad, line 36: shouldn't the naming be existing_users* (plural?)

jadmnsaad (approved):

PR #547, opened 2025-05-28
SUBMIT-541 Change Submission wording - Submitted to EAO

djnunez-aot:


jadmnsaad (approved):

jadmnsaad (changes requested):
I think we should also create a migration to update the logs that were already made in the DB

submit-api/src/submit_api/enums/activity_type.py
```diff
@@ -32,7 +32,7 @@ class VisibilityTypeEnum(enum.Enum):
 class ActivityActionType(enum.Enum):
     """Enum for activity type statuses."""
 
-    ORIGINAL_SUBMISSION = "Original Submission"
+    ORIGINAL_SUBMISSION = "Submitted to EAO"
```
saravanpa-aot: i am not too sure ;but just putting it here ;may be we should the rename the constant name as well if that makes sense as well.. or else we will forget down the line whats the change..
@jadmsaadaot thoughts? 
need not to change if ORIGINAL_SUBMISSION still makes sense

djnunez-aot: We definitely can. I thought they both implied the same thing. We can definitely update the enum itself shouldnt be too big of an issue.

jadmnsaad (approved):

PR #548, opened 2025-05-30
SUBMIT-550 Backend changes to implement the move to folder

VineetBala-AOT:
Ticket: https://eao-dst.atlassian.net/browse/SUBMIT-550

- Added a new end point to handle file move
- Implemented logic to handle version changes and activating old files if a particular version is moved
- Added check to get the correct item id based on the folder to which a move happens


submit-api/src/submit_api/services/submission/submission_creator_factory.py
```diff
@@ -119,6 +123,112 @@ def replace(self, submission_id, request_data):
...
+
+        return new_submission
+
+    def move(self, submission_id, request_data):
+        """Move a document submission."""
+        with session_scope() as session:
+            submission: SubmissionModel = SubmissionModel.find_by_id(submission_id)
+
+            if request_data.get("target_submission_id"):
+                moved_submission = self._create_next_version_of_target(session, submission, request_data)
```
jadmnsaad, line 225: I think we might also nee the "self._restore_previous_active_submission(session, submission)" in the first one. What if we split from a stack and merge onto another

VineetBala-AOT: Thanks, added it.

jadmnsaad (approved):

PR #552, opened 2025-06-03
SUBMIT-507 Letting EAO Create new package versions

djnunez-aot:


submit-api/src/submit_api/resources/staff/package.py
```diff
@@ -85,6 +85,23 @@ def get(original_package_id):
...
+    )
+    @API.response(HTTPStatus.BAD_REQUEST, "Bad Request")
+    @API.response(HTTPStatus.NOT_FOUND, "Not Found")
+    @cors.crossdomain(origin="*")
+    @auth.has_one_of_staff_roles([EpicSubmitRole.EAO_CREATE.value])
+    def post(original_package_id):
+        """Create a new package version."""
+        package_version_data = CreatePackageVersionSchema().load(API.payload)
+        package_with_created_package_version = PackageService.create_new_package_version_with_contacts(
+            package_version_data["package_id"]
```
jadmnsaad: use .get(

submit-api/src/submit_api/schemas/package.py
```diff
@@ -35,6 +35,20 @@ def get_is_approved(obj):
         return obj.package.completed_on is not None
 
 
+class CreatePackageVersionSchema(Schema):
```
saravanpa-aot, line 38: i am not sure if there is another package version schema which can be reused.. since this is just another package , there could be some reusability from the actual package version schema rather than creating a new one..but if there is none, this is good

submit-api/src/submit_api/services/management_plan_service.py
```diff
@@ -133,7 +128,7 @@ def create_new_package_version(cls, item, session):
                 f"Package version not found for item {item.id}.")
         new_package = PackageService.create_new_package_from_original(
             package.id, session)
-        cls._copy_contact_information_from_old_version(package, new_package)
+        PackageUtils.copy_contact_information_from_old_version(package, new_package)
```
saravanpa-aot: I totally love the intention behind doing this..totally appreciative you trying to do it.. it definitely moves us toward better structure and reuse..
But the challenge is its not really a util thing.. its more like another service thing which does many things.. 
Other developers will have a hard time deciding what we will keep in the actual service vs util.. I dont know whats the solution here..we can discuss it..But a util should be for smaller , less impactful stuffs..

submit-api/src/submit_api/services/package.py
```diff
@@ -122,6 +123,45 @@ def _create_package(session, account_project_id, request_data, package_type):
         current_app.logger.info(f"Created package {package.id} for account project {account_project_id}")
         return package
 
+    @classmethod
+    def create_new_package_version_with_contacts(cls, package_id):
+        """Create a new package version with contact information and cleanup."""
+        with session_scope() as session:
```
saravanpa-aot, line 89: Love the readability here

jadmnsaad (approved):

PR #555, opened 2025-06-04
SUBMIT-359, implement Merge/split & documents

jadmnsaad:
- Add a copy s3 object method in document api
- Add move file modal
- Integrate with document api copy object resource
- Integrate with submit api moveSubmission resource


submit-api/migrations/versions/e635c4f73779_rename_folder_values_submitted_documents.py
```diff
@@ -0,0 +1,57 @@
...
+submitted_documents = table(
+    'submitted_documents',
+    column('folder', String)
+)
+
+# Define constants for folder values
+FOLDER_MANAGEMENT_PLAN = 'management_plan'
+FOLDER_MANAGEMENT_PLANS = 'management_plans'
+FOLDER_SUPPORTING = 'supporting'
+FOLDER_SUPPORTING_DOCUMENTS = 'supporting_documents'
```
jadmnsaad, line 29: I did this so the folder names in submitted_document.folder match the folder names in the S3 bucket. Now we use the same const in the frontend to pass the folder name value. Before it was 2

submit-web/src/components/Shared/Table/utils.ts
```diff
@@ -1,4 +1,5 @@
-import { downloadObject } from "@/hooks/api/useObjectStorage";
+import { downloadObject, S3_FOLDER } from "@/hooks/api/useObjectStorage";
+import { camelCase } from "lodash";
```
jadmnsaad, line 2: I didn't love that I'm using lodash, but we are already using this camelCase for the url. So no new lodash code was added. I just moved it to the utils

VineetBala-AOT (approved):

PR #564, opened 2025-06-10
SUBMIT-481 Setup Internal Documents Page

djnunez-aot:


submit-api/migrations/versions/3088bdcf44cc_change_internal_staff_document_to_.py
```diff
@@ -0,0 +1,64 @@
...
+
+    with op.batch_alter_table('items', schema=None) as batch_op:
+        batch_op.drop_constraint('items_package_id_fkey', type_='foreignkey')
+        batch_op.create_foreign_key(None, 'packages', ['package_id'], ['id'])
+
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
```
jadmnsaad, line 46: in the downgrade we didn't repopulate the item_id field

submit-api/migrations/versions/3088bdcf44cc_change_internal_staff_document_to_.py
```diff
@@ -0,0 +1,79 @@
...
+    op.execute("""
+        UPDATE internal_staff_documents isd
+        SET item_id = (
+            SELECT i.id 
+            FROM items i 
+            WHERE i.package_id = isd.package_id 
+            AND i.type_id = (
+                SELECT it.id 
+                FROM item_types it 
+                WHERE it.name = 'EAO Internal Documents'
```
jadmnsaad: there isn't an item type called EAO Internal Documents. I think this code is correct except for the AND part which could just be removed.

jadmnsaad: thinking back on this. I don't think it's actually possible to get the item_id again. Maybe just leave it without repopulating like you initially did it. I'm sorry

djnunez-aot: sounds good reverting now

jadmnsaad (approved):

PR #580, opened 2025-06-17
SUBMIT-555 Setup New status

djnunez-aot:
-Replace NEW_SUBMISSION with NEW

-Still display NEW_SUBMISSION on the eao side in place of SUBMITTED status


submit-api/migrations/versions/fb95dbfcb9d9_add_new_status.py
```diff
@@ -0,0 +1,80 @@
...
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = 'fb95dbfcb9d9'
+down_revision = '3088bdcf44cc'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
jadmsaadaot, line 19: Do we have to add this status in the DB?

jadmnsaad (changes requested):
When will the package ever have a status of NEW in the DB?

jadmnsaad (approved):

PR #640, opened 2025-07-14
SUBMIT-620 Display all packages assigned for collaborator

djnunez-aot:


saravanpa-aot (commented):
i am just wondering if there is a better or simpler way to do this..
or may be write some tests to make sure everything is working well..
if we can avoid fetching from db itself , then we wont have to change the project schema.. or something of that sort...

djnunez-aot:
> i am just wondering if there is a better or simpler way to do this.. or may be write some tests to make sure everything is working well.. if we can avoid fetching from db itself , then we wont have to change the project schema.. or something of that sort...

I will look into another solution

PR #718, opened 2025-12-15
SUBMIT-504:Sorting the project based on name

dinesh-aot:
[https://eao-dst.atlassian.net/browse/SUBMIT-504](https://eao-dst.atlassian.net/browse/SUBMIT-504)

Sort the projects alphabetically on the staff side of submit
Pass claims in the contains_role function in auth to get the checking worked.
image
image

### DIFF
--- a/submit-api/src/submit_api/models/queries/account_project.py
+++ b/submit-api/src/submit_api/models/queries/account_project.py
@@ -115,26 +115,32 @@ class ProjectQueries:
         return query
 
     @classmethod
-    def get_paginated_account_project_ids(cls, account_id: int, filtered_package_ids: list,
-                                          page: int, page_size: int, user: User = None) -> tuple:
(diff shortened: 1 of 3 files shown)


submit-api/src/submit_api/models/queries/account_project.py
jadmnsaad: this operation needs a second and third look. But I think the ordering by name has to happen here is as well in the get_paginated_account_project_ids so the paginated result is also ordered as expected.

jadmnsaad:
Linting issues

submit-api/src/submit_api/models/queries/account_project.py
dinesh-aot: Good catch. I have updated the code. Please check

jadmnsaad (approved):

PR #755, opened 2026-01-09
SUBMIT-697:EAO - Proponents/Holders List

don-aot:
Summary:
This PR adds the Proponents/Holders screen for EAO staff, enabling them to search, filter, sort, and browse all proponents and certificate/exemption holders, and navigate to proponent detail pages.

Changes:
- Added Proponents/Holders to left-side navigation
- Implemented table with Entities, Status, and Actions columns
- Added name search
- Added multi-select Status filter
- Enabled sorting on Entities column
- Added pagination with page size options (10 / 25) and result count display


submit-api/migrations/versions/c7d8e9f0a1b2_create_proponents_table.py
```diff
@@ -0,0 +1,52 @@
...
+    # Create enum type if it doesn't exist (using PostgreSQL-specific syntax)
+    # Using UPPERCASE values to match codebase standard
+    conn = op.get_bind()
+    conn.execute(sa.text("DO $$ BEGIN CREATE TYPE proponentstatus AS ENUM ('ELIGIBLE', 'INELIGIBLE', 'PENDING_ONBOARDING', 'ONBOARDED'); EXCEPTION WHEN duplicate_object THEN null; END $$;"))
+    
+    # Use postgresql.ENUM with create_type=False to prevent SQLAlchemy from trying to create it again
+    proponent_status_enum = postgresql.ENUM('ELIGIBLE', 'INELIGIBLE', 'PENDING_ONBOARDING', 'ONBOARDED', name='proponentstatus', create_type=False)
+    
+    op.create_table('proponents',
+    sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
```
dinesh-aot: We can get rid of the id column

submit-api/migrations/versions/c7d8e9f0a1b2_create_proponents_table.py
```diff
@@ -0,0 +1,52 @@
...
+    sa.Column('proponent_id', sa.Integer(), nullable=False),
+    sa.Column('name', sa.String(), nullable=False),
+    sa.Column('status', proponent_status_enum, nullable=True),
+    sa.Column('is_deleted', sa.Boolean(), nullable=False, server_default='false'),
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('proponent_id')
```
dinesh-aot: The unique key constraint can include is_deleted false

submit-api/src/submit_api/enums/proponent_status.py
```diff
@@ -0,0 +1,13 @@
+"""Proponent status class."""
+from __future__ import annotations
+
+import enum
+
+
+class ProponentStatus(enum.Enum):
+    """Enum for proponent statuses."""
+
+    ELIGIBLE = "ELIGIBLE"
```
dinesh-aot, line 10: Shall we have it like this ELIGIBLE = "Eligible"

submit-api/src/submit_api/resources/staff/proponent.py
```diff
@@ -51,7 +49,31 @@ class Proponents(Resource):
     @cross_origin(origins=allowedorigins())
     def get():
         """Get all proponents."""
```
dinesh-aot: Please put the resource outside staff or proponent folder
/proponents  is enough. No staff or proponent. We are trying to get rid of the staff and proponent segregation for resources
Currently we use project model from proponent service to do proponent related operations. Please check ProjectModel. 
We should move the functionality to the new ProponentModel and adjust the test cases accordingly.

submit-web/src/components/Filters/ProponentStatusFilter.tsx
```diff
@@ -0,0 +1,84 @@
...
+  SelectChangeEvent,
+  Typography,
+} from "@mui/material";
+import { useProponentsHoldersTable } from "@/components/UserManagement/staff/ProponentsHoldersTable/proponentsHoldersTableStore";
+import { ProponentStatus } from "@/models/Proponent";
+import { ProponentStatusChip } from "../ProponentStatusChip";
+import { BCDesignTokens } from "epic.theme";
+
+const PROPONENT_STATUS_OPTIONS: ProponentStatus[] = [
+  "ELIGIBLE",
```
dinesh-aot, line 15: Please check the UI mock up to see if the options are all capital case.

submit-api/src/submit_api/enums/proponent_status.py
don-aot: Initially, I set it as "Eligible", but all other enum values are capitalized (for example, PASSED_CONSULTATION_CHECK = 'PASSED_CONSULTATION_CHECK'). For consistency, we could use the capitalized format.

submit-web/src/components/Filters/ProponentStatusFilter.tsx
don-aot: It won’t be shown in capital letters in the UX. It will appear as follows. Like in the figma

image

submit-api/migrations/versions/c7d8e9f0a1b2_create_proponents_table.py
don-aot: Made the change, we removed the  proponent_id

submit-api/src/submit_api/resources/staff/proponent.py
don-aot: Sure, I will keep the route as it is and create a new /proponents route for backward compatibility.

leodube-aot (dismissed):
Pre-approving, pending updates to address Dinesh's comments

submit-api/migrations/versions/c7d8e9f0a1b2_create_proponents_table.py
don-aot: Done

submit-api/src/submit_api/resources/staff/proponent.py
don-aot: Done

submit-api/src/submit_api/resources/proponent/proponent.py
```diff
@@ -39,19 +38,27 @@
...
     @cross_origin(origins=allowedorigins())
     def get():
-        """Get all proponents."""
-        proponents = ProponentService.get_proponents()
+        """Get all proponents.
+
+        Query parameters:
+        - approved-conditions: If 'true', returns only proponents with approved conditions.
+        """
+        approved_conditions = request.args.get("approved-conditions", "false").lower() == "true"
```
dinesh-aot: Here we might hit on a problem. if we want all proponents to be listed, the by default the value of approved-conditions will be false if do not pass this parameter. So the query will only get proponents with project with approved_condition false where we want all proponents regardless of this value.
It is better to consider None for this parameter if not provided.

submit-web/src/components/ProponentStatusChip/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+type StyleProps = {
+  sx: Record<string, string | number>;
+  label: string;
+};
+
+const statusStyles: Record<ProponentStatus, StyleProps> = {
+  ELIGIBLE: {
+    sx: {
+      borderRadius: "2px",
```
dinesh-aot: Shall we bring the common styles outside to avoid repeating here?

submit-web/src/components/Shared/Table/TablePaginationFooter.tsx
```diff
@@ -0,0 +1,305 @@
...
+  MenuItem,
+  Typography,
+  Link as MuiLink,
+} from "@mui/material";
+import KeyboardDoubleArrowLeftIcon from "@mui/icons-material/KeyboardDoubleArrowLeft";
+import KeyboardDoubleArrowRightIcon from "@mui/icons-material/KeyboardDoubleArrowRight";
+
+const DEFAULT_ROWS_PER_PAGE_OPTIONS = [10, 25] as const;
+
+function getVisiblePages(
```
dinesh-aot: What is the purpose of this component. Submit already has pagination and I think it is using the TablePagination component directly from material design library. Please let me know if it is serving any sort of special purpose

submit-api/tests/unit/resources/test_proponent.py
```diff
@@ -52,32 +57,108 @@ def test_get_proponent_with_projects(client, session):
...
+    assert alpha_proponent["is_deleted"] is False
+
+    beta_proponent = next(p for p in data if p["id"] == proponent2.id)
+    assert beta_proponent["status"] == "ONBOARDED"
+    assert beta_proponent["name"] == "Beta Proponent"
+
+
+def test_get_all_proponents_with_null_status(client, session):
+    """Test get all proponents with null status."""
+    proponent = factory_proponent_model(
```
saravanpa-aot, line 136: id like hard coded number doesnt look great..
```
proponent = factory_proponent_model(
        id=2001,
        name="Null Status Proponent",
        status=None,
        is_deleted=False
    )
```
    
instead of this 

```
proponent = factory_proponent_model(
        name="Null Status Proponent",
        status=None,
        is_deleted=False
    )
    
    proponent_id= proponent.id
```
    
    and let factory_proponent_model create id's within it using Faker ..
Please do it as a follow up PR..dont need to hold this PR for this change

submit-web/src/components/Shared/Table/TablePaginationFooter.tsx
don-aot: This is the common component for the new table footer design. Removed this component and now we are using the default one

submit-web/src/components/ProponentStatusChip/index.tsx
nitheesh-aot: We can just keep the varying parameters in this structure, 
let it be background, borderColor and label

Others like borderRadius, height & border should go in the sx of that Chip.

also would be nice if we use constants for label text

submit-web/src/components/UserManagement/staff/ProponentsHoldersTable/ProponentsHoldersTable.tsx
nitheesh-aot, line 1: is There a common table component we can use for this?

nitheesh-aot (changes requested):

submit-api/tests/unit/resources/test_proponent.py
don-aot: Made the change.

submit-web/src/components/UserManagement/staff/ProponentsHoldersTable/ProponentsHoldersTable.tsx
don-aot: Created a common component for the tables

submit-api/src/submit_api/resources/proponent/proponent.py
don-aot: made the change

submit-web/src/components/ProponentStatusChip/index.tsx
don-aot: made the changes

nitheesh-aot (approved):

PR #757, opened 2026-01-14
SUBMIT-720:Remove proponent_name field from project table and reflect this change everywhere in the app

don-aot:
Summary:
Remove proponent_name from the Project table to avoid data duplication. The proponent name is now accessed through the proponent relationship object, ensuring a single source of truth for proponent information.

Backend Changes:
- Migration: Drop proponent_name column from projects table
- Updated Project model to include proponent object in to_dict()
- Removed proponent_name from ProjectSchema, added nested ProponentSchema
- Updated all test factories and seed functions to remove proponent_name
- Removed redundant Proponent imports in test utilities

Frontend Changes:
- Removed proponent_name from Project TypeScript type
- Added proponent?: Proponent to Project type
- Updated all components to use project.proponent?.name instead of project.proponent_name
- Updated RegistrationUrlCell to use project.proponent_id directly


dinesh-aot (approved):
Looks good to me

PR #772, opened 2026-01-28
Updated Node Version to 24

nitheesh-aot:
Updated node to v24

### DIFF
--- a/submit-web/Dockerfile
+++ b/submit-web/Dockerfile
@@ -1,6 +1,6 @@
 # "build-stage", based on Node.js, to build and compile the frontend
 # pull official base image
-FROM node:20-alpine as build-stage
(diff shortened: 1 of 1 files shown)


submit-web/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # "build-stage", based on Node.js, to build and compile the frontend
 # pull official base image
-FROM node:20-alpine as build-stage
+FROM node:24-alpine as build-stage
```
dinesh-aot, line 3: Is this enough to upgrade or should we check any compatibility issues with the existing packages?

nitheesh-aot: already did that in the local. build and all good

leodube-aot (approved):

PR #778, opened 2026-02-02
SUBMIT-756 Invitation creation refactoring - avoid account project creation upfront

dinesh-aot:
[https://eao-dst.atlassian.net/browse/SUBMIT-756](https://eao-dst.atlassian.net/browse/SUBMIT-756)


leodube-aot (dismissed):

nitheesh-aot (dismissed):

submit-api/src/submit_api/models/account_project.py
```diff
@@ -14,6 +14,9 @@ class AccountProject(BaseModel):
     """Definition of the Account Project entity."""
 
     __tablename__ = 'account_projects'
+    __table_args__ = (
+        UniqueConstraint('account_id', 'project_id', name='uq_account_project'),
+    )
 
     id = Column(db.Integer, primary_key=True, autoincrement=True)
     account_id = Column(db.Integer, ForeignKey('accounts.id', ondelete='CASCADE'), nullable=False)
```
saravanpa-aot, line 22: i forgot if we will have multiple ids in this table..like multiple projects belonging to same proponent?

submit-api/migrations/versions/2656a9b67883_unique_constraint_account_project.py
```diff
@@ -0,0 +1,64 @@
...
+    op.execute("ALTER TYPE submissiontypestatus RENAME TO submissiontype")
+    
+    # Create new submissiontypestatus enum for status column
+    op.execute("CREATE TYPE submissiontypestatus AS ENUM ('SUBMITTED', 'REJECTED', 'APPROVED', 'PENDING')")
+    
+    with op.batch_alter_table('submissions', schema=None) as batch_op:
+        # Change status column from submissionstatus to submissiontypestatus
+        batch_op.alter_column('status',
+               existing_type=postgresql.ENUM('SUBMITTED', 'REJECTED', 'APPROVED', 'PENDING', 'PENDING_REPLACEMENT', name='submissionstatus'),
+               type_=sa.Enum('SUBMITTED', 'REJECTED', 'APPROVED', 'PENDING', name='submissiontypestatus'),
```
saravanpa-aot: what happened to PENDING_REPLACEMENT?

don-aot (dismissed):

submit-api/migrations/versions/2656a9b67883_unique_constraint_account_project.py
saravanpa-aot: i am a bit confused by this enum change..whats the motivation ? can you please add some comments in the PR

submit-api/src/submit_api/services/invitation_service.py
```diff
@@ -62,6 +62,11 @@ def _check_action_authorized(project_ids, account_projects=None, permissions=Non
         else:
             account_project_ids = [ap.id for ap in account_projects]
 
+        # If no account_projects exist yet (new entity invitation not yet accepted),
+        # skip the permission check as it's handled by staff role authorization at API level
+        if not account_project_ids:
+            return
```
saravanpa-aot: i am guessing there wont be any other case like this..like when its new invite , its no account..there wont be any other case without account rite?
i think also we might have to rethink about it now..dont think staff calls this endpoint.. so might be a blank check 
we might have to check if the proponent user actually can see this invitation ..it might not be a staff endpoint

Image

submit-api/src/submit_api/services/invitation_service.py
```diff
@@ -281,14 +289,12 @@ def _get_proponent_by_id(proponent_id):
         return ProponentModel.find_by_id(proponent_id)
 
     @staticmethod
-    def _get_or_create_account_by_proponent(proponent_id, project_ids, session):
+    def _get_or_create_account_by_proponent(proponent_id, session):
```
saravanpa-aot, line 287: so we create account for invitation , but no account projects?

submit-api/migrations/versions/2656a9b67883_unique_constraint_account_project.py
dinesh-aot: image
The above is the current type shows in database.
https://github.com/bcgov/EPIC.submit/pull/778/changes#diff-221c2aaa7ae04a0602da28a8cc383b5255cdf09a0f4028bcffbc7886bf094145
This is the change in the model as per this PR.

dinesh-aot: > what happened to PENDING_REPLACEMENT?

Yeah, missed that. I will add it and update. Thanks

submit-api/src/submit_api/models/account_project.py
dinesh-aot: This model has project_id and account_id as Integers. So if there are multiple projects belongs to the same proponent(Account), we will have multiple entries with different project_id.

submit-api/src/submit_api/services/invitation_service.py
dinesh-aot: > i am guessing there wont be any other case like this..like when its new invite , its no account..there wont be any other case without account rite? i think also we might have to rethink about it now..dont think staff calls this endpoint.. so might be a blank check we might have to check if the proponent user actually can see this invitation ..it might not be a staff endpoint
> 
> Image

When you create invitation for the first time, you will create account. You won't create any account_project till the invited user onboarded.
Yes, you are right about the condition check. This method is used to check permission of the current user on the project for which he will be assigning new user, revoke access, resend invitation. In all these cases we are expecting account project in place prior to the above mentioned actions. Have I addressed the question.
I will remove the checks. Thanks

submit-api/src/submit_api/services/invitation_service.py
dinesh-aot: This method is called when you create an invitation for the first time. mostly for the account admin. We don't have to create account_project during this time. We will still be storing the project_ids in the invitation table for creating the account_project at the time of onboarding.

PR #782, opened 2026-02-03
SUBMIT-691:EAO - View Entity Administrator Contact Information on Proponent/Holder Detail page

don-aot:
Summary:
This PR adds a new Contacts section to the Proponent Details page, allowing EAO staff to view all registered Entity Administrators associated with a proponent. Each administrator is displayed as an individual contact card in a responsive two-column grid layout, providing quick access to essential contact information.

Key Changes:
1. Added the Contacts section below the Projects table on the Proponent Details page
2. Display registered Entity Administrators only (excludes pending/unregistered users)
3. Visible to all EAO users (no permission restrictions)
4. Shows only admins with completed EPIC.submit account registration


nitheesh-aot (dismissed):

submit-api/src/submit_api/models/proponent.py
```diff
@@ -93,4 +101,38 @@ def get_proponent_by_id(cls, proponent_id, include_invitations=False, include_pr
                 "project_id": account_project.project_id,
             } for account_project in account_projects]
 
+        if include_administrators and accounts_ids:
```
saravanpa-aot, line 104: just check what are the permissions for the rest endpoints which calls this ?
the contact is a sensitive information..so just make sure the security is tight

submit-web/src/components/App/Proponents/Contacts/ContactCard.tsx
```diff
@@ -0,0 +1,56 @@
+import { Box, Link, Typography } from "@mui/material";
```
saravanpa-aot, line 1: i almost wish we can do a contact card and pass different params rather than tying it to administrator 
something whcih we can reuse everywhere.. not a must thing..but will be cool to do it that way

saravanpa-aot (commented):
its a good PR.. 
if you handle the auth , you are good to merge..dont have to rewrite the component

PR #783, opened 2026-02-04
SUBMIT-702: EAO - Generating Invite URL - Rules

nitheesh-aot:
[SUBMIT-702](https://eao-dst.atlassian.net/browse/SUBMIT-702): EAO - Generating Invite URL - Rules

[SUBMIT-702]: https://eao-dst.atlassian.net/browse/SUBMIT-702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


submit-api/src/submit_api/services/invitation_service.py
```diff
@@ -441,3 +441,16 @@ def resend_invitation(token):
 
             session.add(invitation)
             return True
+
+    @staticmethod
+    def renew_invitation(invitation_id):
+        """Renew an invitation by ID."""
+        invitation = InvitationsModel.find_by_id(invitation_id)
+        if invitation:
```
leodube-aot, line 452: Should we also confirm if the invitation is PENDING or do you think the frontend check is sufficient?

nitheesh-aot: Good catch @leodube-aot , The invitation status may change to revoked by cron job, which is still in TBD.
I will just add a TODO comment for us to come back to this, once the cron job is decided.

leodube-aot (dismissed):

dinesh-aot (approved):

PR #791, opened 2026-02-09
SUBMIT-739 Update Helm App Title Value to EPIC.submit

leodube-aot:
Ticket: [SUBMIT-739](https://eao-dst.atlassian.net/browse/SUBMIT-739) Fix the name spelling in header > EPIC.submit

**Description**
- Updated Helm app title value to correct spelling. 

@dinesh-aot I'm unfamiliar with Helm so let me know if anything extra is needed for these changes to take effect, or if it will just happen on the next release.

[SUBMIT-739]: https://eao-dst.atlassian.net/browse/SUBMIT-739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deployment/charts/submit-web/values.yaml
+++ b/deployment/charts/submit-web/values.yaml
@@ -4,7 +4,7 @@ app:
   api: https://submit-api-{{ .Release.Namespace }}.apps.gold.devops.gov.bc.ca/api
   env: dev
   version: 1.0.0
-  title: EPIC.Submit
+  title: EPIC.submit
   objectStorageUrl: https://epic-document-{{ .Release.Namespace }}.apps.gold.devops.gov.bc.ca/api
   oidcAuthority: https://dev.loginproxy.gov.bc.ca/auth/realms/eao-epic
   clientId: epic-submit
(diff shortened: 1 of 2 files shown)


saravanpa-aot (changes requested):
do we take it from helm chart ? very less likely
can you check where the value is derived from or may be its just a hard coded text in the UI..

leodube-aot:
> do we take it from helm chart ? very less likely can you check where the value is derived from or may be its just a hard coded text in the UI..

@saravanpa-aot It's not hard coded anywhere. It looks like it's being set as `VITE_APP_TITLE` [here](https://github.com/bcgov/EPIC.submit/blob/develop/submit-web/src/utils/config.ts#L38-L39). I had kind of assumed that the helm chart was setting this variable via the [`configmap.yml`](https://github.com/bcgov/EPIC.submit/blob/develop/deployment/charts/submit-web/templates/configmap.yaml#L14) file, but I could be wrong as I don't fully understand how helm charts work.

How do we set these types of environment variables in deployments? Is it set as a secret in OpenShift?

dinesh-aot (approved):

PR #806, opened 2026-02-27
SUBMIT-768 Track_phases table fixes

dinesh-aot:


submit-api/migrations/versions/f1d8ab8c2d30_adding_display_name_for_phase_name.py
```diff
@@ -0,0 +1,64 @@
...
+down_revision = 'e27054af162b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    # Drop unnecessary indexes from small tables
+    with op.batch_alter_table('package_types', schema=None) as batch_op:
+        batch_op.drop_index('idx_package_types_name')
```
dinesh-aot, line 23: I didn't feel like this index would be useful considering the size possible data in this table.

submit-api/migrations/versions/f1d8ab8c2d30_adding_display_name_for_phase_name.py
```diff
@@ -0,0 +1,64 @@
...
+        batch_op.drop_index('idx_package_types_phase_id')
+
+    # Drop old index and create optimized indexes for packages table
+    with op.batch_alter_table('packages', schema=None) as batch_op:
+        batch_op.drop_index('idx_packages_account_project_work_id')
+    
+    # Create new indexes for packages
+    op.create_index('idx_packages_account_project_id', 'packages', ['account_project_id'])
+    op.create_index(
```
dinesh-aot, line 32: Created this as partial index as there is a chance that this field can contain NULL values in case of Management Plan, which will break the index.

leodube-aot (approved):
Looks good!

My one small concern is the mappings in the `EAActName` and `WorkTypeName` enums. Ideally we could simplify a bit to avoid having to maintain 2 mappings that are inverses of each other. However, since we are not expecting these enums to change frequently I am also good with the implementation as is.

PS: Thanks for the README documentation, will be very helpful.

saravanpa-aot (approved):

PR #831, opened 2026-03-24
SUBMIT-781 Create backend unit tests for models and updated invitation logic

leodube-aot:
Ticket: [SUBMIT-781](https://eao-dst.atlassian.net/browse/SUBMIT-781) Create backend unit tests for work/phase/invitation/account management 

**Description**
- Created unit tests for all the model files (that have ORM operations).
- Updated existing unit tests that needed updating (specifically regarding the latest changes around invitations, work+phase, and account management).
- Did some minor refactoring to minimize deprecation warnings, and to keep ORM operations out of the services.

**Refactoring**
- Fixed a bunch of deprecation warnings. The backend tests used to throw 228 warnings. Now they only throw 2 warnings. Most of these were fixed by updating `datetime.utcnow` to `datetime.now(UTC)`.
- Moved some methods out of `/services` into specific `/models` files. This was to try to keep direct orm operations in the models (ie. `db.query` operations)
- Create a `persist` method on the `BaseModel` to avoid duplicate code.

[SUBMIT-781]: https://eao-dst.atlassian.net/browse/SUBMIT-781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


submit-api/src/submit_api/models/queries/package.py
```diff
@@ -126,7 +126,6 @@ def aggregate_item_statuses(cls, items: list):
         cls._add_completed_status(aggregated_statuses, statuses)
         cls._add_submitted_status(aggregated_statuses, statuses)
         cls._add_passed_consultation_check(aggregated_statuses, statuses)
-        cls._add_review_rejected(aggregated_statuses, statuses)
```
leodube-aot, line 135: duplicated from line 122

submit-api/src/submit_api/models/account.py
```diff
@@ -32,8 +32,10 @@ def create_account(cls, account_data, session=None) -> Account:
         account = Account(
             proponent_id=account_data.get('proponent_id', None),
         )
-        if session:
-            account.flush()
-        else:
-            account.save()
-        return account
+        return account.persist(session)
```
leodube-aot, line 35: We use this pattern quite frequently: if session - flush, else - save. I moved this into a common method called `persist` that does the same thing, but avoids duplicate code.

submit-api/src/submit_api/schemas/package.py
```diff
@@ -138,8 +138,7 @@ class Meta:  # pylint: disable=too-few-public-methods
     name = fields.Str(data_key="name")
     type = fields.Nested(PackageTypeSchema, data_key="type")
     type_id = fields.Int(data_key="type_id")
-    status = fields.List(fields.Enum(enum=PackageStatus),
-                         enum=PackageStatus, data_key="status")
+    status = fields.List(fields.Enum(enum=PackageStatus), data_key="status", metadata={"enum": PackageStatus})
```
leodube-aot, line 141: Marshmallow 3.x deprecated passing extra metadata (like enum, description, example, etc.) as direct keyword arguments to fields. They need to be wrapped in an explicit metadata={} dict.

submit-api/src/submit_api/services/account_user_service.py
```diff
@@ -104,18 +102,12 @@ def _fetch_package_names(original_package_ids: list[int]) -> dict[int, str]:
...
-            )
-        return query.all()
+        return AccountUserModel.get_filtered_by_account_id(account_id, account_project_ids)
 
     @staticmethod
     def _fetch_user_status_name(user_id):
         """Fetch only the user's status name."""
-        user = UserModel.query.filter(UserModel.id == user_id).first()
-        return user.user_status.status_name if user else None
+        return UserModel.get_status_name_by_id(user_id)
```
leodube-aot, line 110: Part of the refactoring: Encapsulated `Model.query.filter` calls and moved into the model file itself. Now we call the encapsulated query. No logic actually changes here.

submit-api/src/submit_api/models/proponent.py
```diff
@@ -52,87 +46,6 @@ def get_all_proponents(cls, include_deleted=False, approved_conditions_only=None
         return query.order_by(cls.name).all()
 
     @classmethod
-    def get_proponent_by_id(
```
leodube-aot, line 55: This wasn't deleted, just moved into the `proponent_service` file. This method does a lot of orchestration work, pulling from several other db models. Makes more sense as a service.

nitheesh-aot (dismissed):

nitheesh-aot (approved):

PR #856, opened 2026-04-16
SUBMIT-761 Additional Information Submission - IN PROGRESS status for package

dinesh-aot:
Added IN_PROGRESS Status for work related submissions. The logic is as follows
The work related submissions doesn't have to go through the process of management plan instead it can be IN_PROGRESS as soon as it is created and be in that status until it is SUBMITED TO EAO. When Submitted to EAO, the status should change to SUBMITTED.


nitheesh-aot (dismissed):
LGTM

lint errors to be fixed

PR #862, opened 2026-04-20
SUBMIT-771: EAO - Document Library - View Documents for Selected Project

nitheesh-aot:
image

### DIFF
--- a/submit-api/src/submit_api/resources/submitted_document.py
+++ b/submit-api/src/submit_api/resources/submitted_document.py
@@ -22,7 +22,8 @@ from flask_restx import Namespace, Resource
 from submit_api.auth import auth
 from submit_api.models.account_project_search_options import DocumentSearchOptions, ProjectDocumentSearchOptions
 from submit_api.resources.apihelper import Api as ApiHelper
-from submit_api.schemas.submission import SubmissionSchema, SubmittedDocumentByProjectSchema, PaginatedProjectDocumentItemSchema
+from submit_api.schemas.submission import (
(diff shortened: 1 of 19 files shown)


submit-api/src/submit_api/resources/submitted_document.py
```diff
@@ -56,6 +57,26 @@ class AccountDocuments(Resource):
     def get():
         """Get all submitted documents."""
         args = request.args
+        project_id = args.get('project_id', type=int)
+
+        if project_id:
+            page = args.get('page', 1, type=int)
```
dinesh-aot, line 63: Do we have to specify the explicit type like this. I think it will automatically cast from what is coming through the arguments.

submit-api/src/submit_api/resources/submitted_document.py
dinesh-aot, line 86: It would be great to avoid two types being returned from one endpoint if manageable.

submit-api/src/submit_api/schemas/submission.py
```diff
@@ -98,3 +98,32 @@ class Meta:  # pylint: disable=too-few-public-methods
...
+
+class PaginatedProjectDocumentItemSchema(Schema):
+    """Paginated project document item schema."""
+
+    class Meta:  # pylint: disable=too-few-public-methods
+        """Exclude unknown fields in the deserialized output."""
+
+        unknown = EXCLUDE
+
+    id = fields.Int(data_key="id")
```
dinesh-aot, line 111: please check if this data_key is required or not. Because the field name is same as that of what you mentioned in data_key. Also, put a metadata={"description": "The description of the field"}. This will be useful for swagger documentation

dinesh-aot (approved):

PR #867, opened 2026-04-22
SUBMIT-790 Send Update Request to Proponent

dinesh-aot:
The primary purpose of this PR is to address the requirements in the below mentioned ticket.
[https://eao-dst.atlassian.net/browse/SUBMIT-790](https://eao-dst.atlassian.net/browse/SUBMIT-790)
I have handled some of the items mentioned in the refactor tickets as well.
[https://eao-dst.atlassian.net/browse/SUBMIT-823](https://eao-dst.atlassian.net/browse/SUBMIT-823)
version_enabled, mandatory fields are added on package_type level.
[https://eao-dst.atlassian.net/browse/SUBMIT-824](https://eao-dst.atlassian.net/browse/SUBMIT-824)
is_required field is added on the packate_item_type mapping table to configure which item is required or not at various package_types


submit-api/src/submit_api/models/queries/package.py
```diff
@@ -18,29 +18,51 @@
...
 from submit_api.models.package_version import PackageVersion
+from submit_api.models.package_item_type import PackageItemType
 
 
 # pylint: disable=too-few-public-methods
 class PackageQueries:
     """Query module for complex package queries"""
 
+    @classmethod
+    def _get_required_item_statuses(cls, items: list) -> list[str]:
```
dinesh-aot, line 29: These changes are in relation with is_required field added on the package_type_item level. By default all items will be required. We can flag them off selectively for our requirement. 
The status check for COMPLETE is now check only items with is_required = true

submit-api/src/submit_api/__init__.py
```diff
@@ -21,9 +21,9 @@
 csp = (
     secure.ContentSecurityPolicy()
     .default_src("'self'")
-    .script_src("'self' 'unsafe-inline'")
-    .style_src("'self' 'unsafe-inline'")
-    .img_src("'self' data:")
+    .script_src("'self'", "'unsafe-inline'")
+    .style_src("'self'", "'unsafe-inline'")
+    .img_src("'self'", "data:")
```
leodube-aot, line 26: Thanks for fixing this, I was wondering why the CI tests were failing. I'm guessing one of the package version bumps caused the issue.

leodube-aot (approved):
Looks good, the new fields on package_types and package_item_types will be handy going forward.

nitheesh-aot (approved):

PR #868, opened 2026-04-22
SUBMIT-782 Use isWorkRelated flag rather than explicit submission type checks

leodube-aot:
Description
- This was part of my upcoming PR for SUBMIT-788, but it fixes a bug in how eligible work related projects are displayed in the Proponents/Holders table.
- Pulled it out of the bigger PR in order to get merged quicker and unblock MA.

### DIFF
--- a/submit-api/src/submit_api/schemas/project.py
+++ b/submit-api/src/submit_api/schemas/project.py
@@ -26,7 +26,9 @@ class ProjectSchema(Schema):
     proponent = fields.Nested(ProponentSchema, data_key="proponent", allow_none=True)
     ea_certificate = fields.Str(data_key="ea_certificate")
     epic_guid = fields.Str(data_key="epic_guid")
+    has_approved_condition = fields.Boolean(data_key="has_approved_condition")
     works = fields.List(fields.Nested(TrackWorkSchema, data_key="works"))
+    current_work = fields.Nested(TrackWorkSchema, data_key="current_work")
 
 
 class AddProjectSchema(Schema):
@@ -65,6 +67,7 @@ class AccountProjectSchema(Schema):
     project = fields.Nested(ProjectSchema, data_key="project")
     latest_packages = fields.List(fields.Nested(AccountProjectPackageSchema), data_key="packages")
     account_project_works = fields.List(fields.Nested(AccountProjectWorkSchema, data_key="account_project_works"))
+    is_work_related = fields.Boolean(data_key="is_work_related")
 
 
 class StaffAccountProjectPackageSchema(StaffPackageSchema):
--- a/submit-web/src/components/App/Projects/Project.tsx
+++ b/submit-web/src/components/App/Projects/Project.tsx
@@ -10,10 +10,10 @@ type ProjectParam = {
 
 export const Project = ({ accountProject }: ProjectParam) => {
   const navigate = useNavigate();
-  const { name, ea_certificate } = accountProject.project;
-  const currentWork =
-    accountProject.account_project_works?.at(-1)?.work ?? null;
-  const currentPhase = currentWork?.current_phase ?? null;
+  const { name, ea_certificate, current_work } = accountProject.project;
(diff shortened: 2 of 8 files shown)


submit-api/src/submit_api/schemas/project.py
```diff
@@ -26,7 +26,9 @@ class Meta:  # pylint: disable=too-few-public-methods
     proponent = fields.Nested(ProponentSchema, data_key="proponent", allow_none=True)
     ea_certificate = fields.Str(data_key="ea_certificate")
     epic_guid = fields.Str(data_key="epic_guid")
+    has_approved_condition = fields.Boolean(data_key="has_approved_condition")
     works = fields.List(fields.Nested(TrackWorkSchema, data_key="works"))
+    current_work = fields.Nested(TrackWorkSchema, data_key="current_work")
```
dinesh-aot, line 31: Why would we need this field.

submit-web/src/components/App/Projects/Project.tsx
```diff
@@ -10,10 +10,10 @@ type ProjectParam = {
...
 export const Project = ({ accountProject }: ProjectParam) => {
   const navigate = useNavigate();
-  const { name, ea_certificate } = accountProject.project;
-  const currentWork =
-    accountProject.account_project_works?.at(-1)?.work ?? null;
-  const currentPhase = currentWork?.current_phase ?? null;
+  const { name, ea_certificate, current_work } = accountProject.project;
+  const status =
+    current_work?.current_phase?.name || PROJECT_STATUS.POST_DECISION;
+  const title = current_work?.title || "Management Plans & Related Documents";
```
dinesh-aot, line 16: I actually wanted to talk to you about this. Shall we consider this a list of works and expect to show multiple submission cards. Because eventually we will be showing multiple works such as Extension and Amendments together under the main Project Container.

leodube-aot: So we should iterate over the list of works and display the `ProjectSubmissionsCard` component for each work? Do we need to consider whether the work is IN_PROGRESS?

submit-api/src/submit_api/schemas/project.py
leodube-aot: I was aiming to simplify some of the frontend computation around finding the current_work (for example at [your comment below](https://github.com/bcgov/EPIC.submit/pull/868/changes#diff-269a27899936544261b3508e2d7f8673148a64d5047417e337c78b0ee7aa28c6L14-R15)). The frontend check wasn't considering if the work was active or not-deleted or IN_PROGRESS, so rather than calculate that in the frontend by iterating over the account_project_works list I included it in the API response.

It's still doable to do in the frontend, and I can update it if you think that makes more sense

submit-web/src/components/App/Projects/Project.tsx
dinesh-aot: When you actually onboard a work through the below screen, we should show only work with state IN_PROGRESS and the current_phase of that should be enabled for submit.
image
Then to see if the package is work related, we can check if the package has associated account_project_work_id.

leodube-aot: Yup, there's a ticket for correcting the Eligible Projects table: https://eao-dst.atlassian.net/browse/SUBMIT-836. For projects with `has_approved_conditions`, but that has ineligible work/phase, will display Management Plan & Related Documents - Post Decision.

So you're saying, for the purposes of this PR:
- The is no need to have an explicit is_work_related flag in the backend or frontend, as we'll check account_project_work_id.

dinesh-aot: yes

leodube-aot:
Closing as https://github.com/bcgov/EPIC.submit/pull/871 supercedes it. In the new PR I removed the `is_work_related` flag from the API in favor of checking for an `account_project_work`.

PR #875, opened 2026-04-29
SUBMIT-789 EAO - Acknowledge IPD Documents & Submission

leodube-aot:
Ticket: [SUBMIT-789](https://eao-dst.atlassian.net/browse/SUBMIT-789) EAO - Acknowledge IPD Documents & Submission

**Description**
_Frontend_
- Updated `DocumentRow` component to work for acknowledgements.
- Refactored `StatusCell` component to make it more data driven.

_Backend_
- Created a declarative transition map to enforce Type A/B/C state transitions.

**Screenshots**

Screenshot 2026-04-29 at 8 59 04 AM

[SUBMIT-789]: https://eao-dst.atlassian.net/browse/SUBMIT-789?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/submit-api/src/submit_api/models/submission.py
+++ b/submit-api/src/submit_api/models/submission.py
@@ -37,21 +37,22 @@ class SubmissionStatus(enum.Enum):
 # and what the only allowed "next" status is from each state.
 ALLOWED_TRANSITIONS: dict[PackageApprovalType, dict[SubmissionStatus, set[SubmissionStatus]]] = {
     PackageApprovalType.A: {
-        SubmissionStatus.SUBMITTED:  {SubmissionStatus.VERIFIED},
+        SubmissionStatus.SUBMITTED: {SubmissionStatus.VERIFIED},
         SubmissionStatus.VERIFIED: {SubmissionStatus.SUBMITTED},
     },
     PackageApprovalType.B: {
-        SubmissionStatus.SUBMITTED:      {SubmissionStatus.VERIFIED, SubmissionStatus.ACKNOWLEDGED},
-        SubmissionStatus.VERIFIED:     {SubmissionStatus.SUBMITTED, SubmissionStatus.ACKNOWLEDGED},
+        SubmissionStatus.SUBMITTED: {SubmissionStatus.VERIFIED, SubmissionStatus.ACKNOWLEDGED},
+        SubmissionStatus.VERIFIED: {SubmissionStatus.SUBMITTED, SubmissionStatus.ACKNOWLEDGED},
         SubmissionStatus.ACKNOWLEDGED: {SubmissionStatus.VERIFIED},
     },
     PackageApprovalType.C: {
-        SubmissionStatus.SUBMITTED:      {SubmissionStatus.VERIFIED, SubmissionStatus.ACKNOWLEDGED},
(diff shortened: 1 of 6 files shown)


submit-api/src/submit_api/models/submission.py
```diff
@@ -31,6 +32,27 @@ class SubmissionStatus(enum.Enum):
...
+        SubmissionStatus.ACKNOWLEDGED: {SubmissionStatus.VERIFIED},
+    },
+    PackageApprovalType.C: {
+        SubmissionStatus.SUBMITTED:      {SubmissionStatus.VERIFIED, SubmissionStatus.ACKNOWLEDGED},
+        SubmissionStatus.VERIFIED:     {SubmissionStatus.SUBMITTED, SubmissionStatus.ACKNOWLEDGED},
+        SubmissionStatus.ACKNOWLEDGED: {SubmissionStatus.VERIFIED, SubmissionStatus.APPROVED},
+        SubmissionStatus.APPROVED:     {SubmissionStatus.ACKNOWLEDGED},
+    },
+}
+
```
leodube-aot, line 55: The goal here is to avoid a bunch of `if-else` checks in validating what state transitions are valid. For example, a type A submission can move from SUBMITTED to VERIFIED, and also in reverse. Type B and type C have more advanced state transitions that are handled by this mapping.

submit-api/src/submit_api/services/submission/__init__.py
```diff
@@ -121,13 +122,47 @@ def update_submission_status(cls, submission_id, status):
             submission = cls.get_submission_by_id(submission_id)
             existing_status = submission.status
             if existing_status != status:
+                # Resolve the approval type from the submission's package
+                package = cls.get_package_by_submission_id(submission_id)
+                approval_type = package.type.approval_type
+
+                if approval_type is None:
+                    raise ValueError(
```
dinesh-aot, line 130: We should make sure how we are handling the existing data and package types. Like Management Plan. Because does it follow our Type methodology?

leodube-aot: Looking at the Submission Types flow, it looks like MP falls under Type C? Which is confusing since it has a completely different flow right now on the staff side. Maybe a question for MA when she returns.

I personally think this endpoint should ONLY handle submissions that fall under our Type methodology. The existing MP flow does not use this endpoint, so I don't really see a reason to update it right now.

 
Screenshot 2026-04-29 at 1 45 44 PM

dinesh-aot: > Looking at the Submission Types flow, it looks like MP falls under Type C? Which is confusing since it has a completely different flow right now on the staff side. Maybe a question for MA when she returns.
> 
> I personally think this endpoint should ONLY handle submissions that fall under our Type methodology. The existing MP flow does not use this endpoint, so I don't really see a reason to update it right now.
> 
> Screenshot 2026-04-29 at 1 45 44 PM

Its fine if the MP doesn't use this endpoint.

dinesh-aot (approved):

PR #879, opened 2026-05-07
SUBMIT-817 EAO - Approve IPD Submission & Refactoring

leodube-aot:
Ticket: [SUBMIT-817](https://eao-dst.atlassian.net/browse/SUBMIT-817) EAO - Approve IPD Submission

**Description**
_Frontend_
- Refactored `PackageStatusChip` and `SubmissionStatusChip` to minimize duplicate code. All of the chips have some common ui styling, and can be categorized into themes. This makes adding and managing statuses much easier going forward.
- Refactored `DocumentRow` and the staff submission page (`.../_submissionLayout/index.tsx`). There was lots of complexity pilling up here, so I split most of the business logic into hooks.
  - `useStaffSubmissionPage`
  - `useUpdateRequests`
  - `useDocumentRow`

_Backend_
- Refactored `PackageQueries` into 3 different classes:
  - `PackageQueries`: handles general queries
  - `PackageItemQueries`: handles package status operations that are driven by aggregating **item** statuses. Most of the existing logic now lives here.
  - `PackageSubmissionQueries`: handles package status operations that are driven by aggregating **submission** statuses. Most of the new logic added in this PR lives here, and should be used for the Type A/B/C staged workflows.
- Added `PackageService.approve_package`

_Database_
- Added several enum values to the `packagestatus` enum.

[SUBMIT-817]: https://eao-dst.atlassian.net/browse/SUBMIT-817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


submit-api/src/submit_api/models/queries/package.py
```diff
@@ -179,102 +236,148 @@ def aggregate_item_statuses(cls, items: list):
...
         cls._add_under_cc_check(aggregated_statuses, all_statuses)
         cls.add_awaiting_manager_review(aggregated_statuses, all_statuses)
-        cls._add_additional_info_status(aggregated_statuses, items)
 
         aggregated_statuses_list = list(aggregated_statuses)
         return aggregated_statuses_list
 
-    @classmethod
-    def _add_additional_info_status(cls, aggregated_statuses: set, items: list):
-        """Find Additional Information package status based on document verification."""
```
leodube-aot, line 189: Removed this since `update_package_status_from_submissions` should now correctly handle statuses for Additional Information packages.

submit-api/src/submit_api/schemas/package.py
```diff
@@ -211,8 +211,14 @@ def map_status(self, data, many, **kwargs):
...
-        new_status = set(status for status in new_status if status)
-        data['status'] = list(new_status)
+        # Return new status list while preserving insertion order
+        seen = set()
+        deduped_status = []
+        for status in new_status:
+            if status and status not in seen:
+                seen.add(status)
+                deduped_status.append(status)
+        data['status'] = deduped_status
```
leodube-aot, line 221: Converting to a `set` destroys ordering, and then converting back to a `list` produces an unpredictable order (often alphabetical-ish, but not guaranteed).

This preserves insertion order — the database statuses come through first (in their original order), followed by any appended non-canonical statuses like `UPDATE_REQUESTED` or `REVISION_REQUIRED`.

submit-api/src/submit_api/models/queries/package.py
```diff
@@ -11,21 +11,78 @@
...
+                func.array_agg(
+                    func.json_build_object(
+                        "id", PackageModel.id,
+                        "name", PackageModel.name,
+                        "original_package_id", PackageVersion.original_package_id,
+                    )
+                ).label("packages")  # Aggregate packages as a JSON array
+            )
+            .join(PackageModel, PackageModel.account_project_id == AccountProject.id)
+            .join(PackageVersion,
```
dinesh-aot, line 62: Shall we use leftoin here as for some packages(enabling_version set to false) there might not be entries in package version table

leodube-aot: Yeah I think that makes sense. This is actually quite an old method (I didn't update anything here, just moved it around in the file) so the logic might be a bit outdated with all the updates we've been doing.

I'll take a look and update this join if it checks out

dinesh-aot (approved):

submit-api/src/submit_api/models/queries/package.py
leodube-aot: I tracked down where this code is used and it's for populating the submissions in the New User Form. It definitely needs to be fixed since I also have an Additional Information submission on the project that is not appearing.

A simple left join didn't quite fix the issue. Since it'll need a bit more work, I'll either fix in my next ticket, or I'll create a bug ticket.

Screenshot 2026-05-07 at 9 45 47 AM

PR #887, opened 2026-05-11
SUBMIT-828  Upload Updated Documents and Resubmit Package to EAO

dinesh-aot:
[https://eao-dst.atlassian.net/browse/SUBMIT-828](https://eao-dst.atlassian.net/browse/SUBMIT-828)


submit-api/src/submit_api/services/package_service.py
```diff
@@ -456,10 +457,6 @@ def _submit_package(cls, package, session):
     def _resubmit_package(cls, package, session):
         """Submit the package by updating its status and items."""
         current_app.logger.info(f"Resubmitting package {package.id}")
-        open_update_requests = [request for request in package.update_requests
-                                if request.status != UpdateRequestStatus.ACCEPTED.value]
-        if not open_update_requests:
```
dinesh-aot, line 461: This validation is already happening at higher level. See line 303 in package_service.py

submit-api/src/submit_api/models/base_model.py
```diff
@@ -30,8 +30,8 @@ class BaseModel(db.Model):
 
     __abstract__ = True
 
-    created_date = db.Column(db.DateTime, default=datetime.now(UTC), nullable=False)
-    updated_date = db.Column(db.DateTime, default=datetime.now(UTC), onupdate=datetime.now(UTC), nullable=True)
+    created_date = db.Column(db.DateTime, default=lambda: datetime.now(UTC), nullable=False)
+    updated_date = db.Column(db.DateTime, default=lambda: datetime.now(UTC), onupdate=lambda: datetime.now(UTC), nullable=True)
```
dinesh-aot: these date fields were getting the same timestamp when the module was initialized before. Now it will capture the correct value up on each time this is called.

PR #918, opened 2026-06-04
SUBMIT-836 Eligibility Entries added while onboarding proponent

dinesh-aot:
[https://eao-dst.atlassian.net/browse/SUBMIT-836](https://eao-dst.atlassian.net/browse/SUBMIT-836)
Now we can see the eligible entries like this 
image
All the works + non works will be listed.
We have account_project_works table to have granular control over the account project and work. I have in introduced a new type for nonworkitems and a table to map the account project and non work item.

The invitation table has a column called project_ids which will only indicate which are all the project the invited user can access not at a Management Plan level or work level. So I have added a new column called project_selections which is a json type and will store the project_id and the possible works and non_worktypes.

I have also handled the existing invitations using a migration.
Please check the below eligible entries when the proponent onboards
image
Now you will see each work or non work for the given project in a separate card. Instead of calling this a project list, I call them Eligible Entries.


submit-api/migrations/versions/861b5e7374e0_nonwork_item_related_changes.py
```diff
@@ -0,0 +1,83 @@
...
+    sa.Column('created_date', sa.DateTime(), nullable=False),
+    sa.Column('updated_date', sa.DateTime(), nullable=True),
+    sa.Column('created_by', sa.String(length=50), nullable=True),
+    sa.Column('updated_by', sa.String(length=50), nullable=True),
+    sa.ForeignKeyConstraint(['account_project_id'], ['account_projects.id'], ondelete='CASCADE'),
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('account_project_id', 'non_work_item_type', name='uq_account_project_non_work')
+    )
+    with op.batch_alter_table('invitations', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('project_selections', postgresql.JSON(astext_type=sa.Text()), nullable=True))
```
saravanpa-aot: project_selections means the project they got inivted to rite?
may be just a name change as invited_projects or projects_invited might be easier to read

saravanpa-aot (dismissed):

PR #966, opened 2026-07-13
SUBMIT-883-Refactoring on the account-projects filtering logic and permission related fixes

dinesh-aot:
Please check the account_project.py query file primarily as there were a lot of code I commented to make it simpler. Previously we were querying package_ids and project_ids separately and then query the projects and there were a lot redundant code was there. I just cleaned it up and tested it with all the roles we have except the CONTRIBUTOR role of the proponent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -168,4 +168,5 @@ deployment/charts/submit-web/values.prod.yaml
 deployment/charts/submit-api/values.prod.yaml
 deployment/charts/submit-api/values.dev.yaml
 deployment/charts/submit-web/values.dev.yaml
+.windsurf/
 .kiro/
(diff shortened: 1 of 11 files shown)


leodube-aot (dismissed):
Makes sense for the most part. Once merged I'll do a bit of testing as well in DEV.

Edit: couple failing unit tests that might need to be fixed.

nitheesh-aot:
looks good to me, 
can we remove the code rather than commenting them out? @dinesh-aot

dinesh-aot:
> looks good to me, can we remove the code rather than commenting them out? @dinesh-aot

Sure, I will do that

dinesh-aot:
> looks good to me, can we remove the code rather than commenting them out? @dinesh-aot

I will fix those lint and test issues and merge. Thanks

PR #969, opened 2026-07-14
SUBMIT-896 Handle unfinished submissions with required documents

leodube-aot:
Tickets: [SUBMIT-896](https://eao-dst.atlassian.net/browse/SUBMIT-896) Error message when trying to "save & exit" after deleting the only file in the section.

**Description**
- Previously we were trying to mark partially completed submissions as COMPLETED, which was throwing the API error. Instead, we should check if the required files have been uploaded, and if not then when we save the submission we pass it's current status.

[SUBMIT-896]: https://eao-dst.atlassian.net/browse/SUBMIT-896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/submit-web/src/components/App/SubmissionItem/AdditionalInformation/ProponentView/index.tsx
+++ b/submit-web/src/components/App/SubmissionItem/AdditionalInformation/ProponentView/index.tsx
@@ -94,10 +94,7 @@ export const AdditionalInformationProponentView = () => {
 
   const handleCompleteForm = (formData: AdditionalInformationForm) => {
     if (!formData.uploadDocuments?.length) {
-      saveSubmission(
-        formData,
-        SUBMISSION_ITEM_STATUS.PARTIALLY_COMPLETED.value,
-      );
+      saveSubmission(formData, submissionItem?.status);
     } else {
       saveSubmission(formData, SUBMISSION_ITEM_STATUS.COMPLETED.value);
     }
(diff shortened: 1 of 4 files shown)


submit-web/src/components/App/SubmissionItem/AdditionalInformation/ProponentView/index.tsx
```diff
@@ -89,12 +93,16 @@ export const AdditionalInformationProponentView = () => {
   });
 
   const handleCompleteForm = (formData: AdditionalInformationForm) => {
-    saveSubmission(formData, SUBMISSION_ITEM_STATUS.COMPLETED.value);
+    if (!formData.uploadDocuments?.length) {
+      saveSubmission(formData, submissionItem?.status);
+    } else {
```
dinesh-aot, line 98: Pleaes check if this is applicable to other management plan components as lot of the items are marked as required in the backend. 
Plus, it would  be great to check if the item is required by checking the is_required flag before sending the status

leodube-aot: This doesn't affect management plan related documents since we have 2 separate action buttons: 
- Save & Continue Later (does not care if required document is there).
- Save Completed Form (does check for required document, which we want it to do here).
Screenshot 2026-07-17 at 10 20 49 AM

I'll look into if we can use the `is_required` flag, but it might be a bit tricky because the `formData` object stores the document types as attributes. Comparing `SubmissionItemType` to the `formData` object might not be possible without adding additional complexity (ie. some sort of mapping), but I'll double check.

dinesh-aot (approved):

PR #1002, opened 2026-07-29
SUBMIT-896 Fix removal of early engagement submission files

leodube-aot:
Tickets: [SUBMIT-896](https://eao-dst.atlassian.net/browse/SUBMIT-896) Error message when trying to "save & exit" after deleting the only file in the section.

**Issue**
- Previously we updated the submission to allow items to "Save & Exit" when no required files are present. An edge case was still throwing an error when a user adds a file, clicks "Save & Exit", and then returns to the submission item and removes the file and then tries to "Save & Exit" again.
- The cause of the bug is that on the first save, the submission has a status of `COMPLETED`. When we remove the file, we were passing the previous submission status of `COMPLETED`, however now it was missing the required document.

**Fix**
- When a submission has a status of `COMPLETED`, but is missing a required doc, we pass `PARTIALLY_COMPLETED`.

[SUBMIT-896]: https://eao-dst.atlassian.net/browse/SUBMIT-896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/submit-web/src/components/App/SubmissionItem/AdditionalInformation/ProponentView/index.tsx
+++ b/submit-web/src/components/App/SubmissionItem/AdditionalInformation/ProponentView/index.tsx
@@ -94,7 +94,12 @@ export const AdditionalInformationProponentView = () => {
 
   const handleCompleteForm = (formData: AdditionalInformationForm) => {
     if (!formData.uploadDocuments?.length) {
-      saveSubmission(formData, submissionItem?.status);
+      saveSubmission(
+        formData,
+        submissionItem?.status != SUBMISSION_ITEM_STATUS.COMPLETED.value
+          ? submissionItem?.status
+          : SUBMISSION_ITEM_STATUS.PARTIALLY_COMPLETED.value,
+      );
     } else {
       saveSubmission(formData, SUBMISSION_ITEM_STATUS.COMPLETED.value);
     }
--- a/submit-web/src/components/App/SubmissionItem/EPSubmission/EPProponentView/index.tsx
+++ b/submit-web/src/components/App/SubmissionItem/EPSubmission/EPProponentView/index.tsx
@@ -103,7 +103,9 @@ export const EngagementPlanProponentView = () => {
     if (!formData.engagementPlan?.length) {
       saveSubmission(
         formData,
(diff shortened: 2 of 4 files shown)


dinesh-aot (approved):

submit-web/src/components/App/SubmissionItem/AdditionalInformation/ProponentView/index.tsx
```diff
@@ -94,7 +94,12 @@ export const AdditionalInformationProponentView = () => {
 
   const handleCompleteForm = (formData: AdditionalInformationForm) => {
     if (!formData.uploadDocuments?.length) {
-      saveSubmission(formData, submissionItem?.status);
+      saveSubmission(
+        formData,
+        submissionItem?.status != SUBMISSION_ITEM_STATUS.COMPLETED.value
```
dinesh-aot, line 99: Can you check if we use PARTIALLY_COMPLETED status at all for these submissions. 
When you create a submission by default it is PENDING. 
Then, when you submit the document, it will become COMPLETED. 
So, if you have to revert it, should it be PENDING. Please check

leodube-aot: This status is for the submission **item** status (in the db: `items.status`) which doesn't have PENDING as one of it's options (see [here](https://github.com/bcgov/EPIC.submit/blob/develop/submit-web/src/models/Submission.ts#L29-L54) for available item statuses). The `PENDING` status is for the submissions (db: `submissions.status`) which stays as `PENDING` even if we mark the item as `PARTIALLY_COMPLETED` here.

We use `PARTIALLY_COMPLETED` as an item status in Management Plans when a user clicks `Save & Exit` on an uncompleted submission item (see [here](https://github.com/bcgov/EPIC.submit/blob/develop/submit-web/src/components/App/SubmissionItem/ManagementPlanSubmission/ManagementPlanProponentView/index.tsx#L200)).

I could be open to using a different `SubmissionItemStatus` for this use case (or even creating a new one), but I think `PARTIALLY_COMPLETED` makes sense syntactically with existing logic (ie. Management Plans).

